PR #34, opened 2023-06-13
[feat] support refreshing in error state

manuel-plavsic:
This PR adds support for refreshing while in error state. Docs and tests have been updated to make use of this functionality.

If one does not want to keep displaying the error while refreshing, I guess they can create a custom function, e.g. `displayLoading`, and call it both in the `ready` and `loading` callbacks, e.g.:

```dart
ResourceBuilder(
  resource: user,
  builder: (_, userValue) {
    Widget displayLoading() {
      // ...
    }
  
    return userValue.on(
      ready: (data, refreshing) {
        if (refreshing) {
          // ... display data with CircularProgressIndicator
        } else {
          // ...
        }
      },
      error: (error, _, refreshing) {
        if (refreshing) {
          return displayLoading();
        } else {
          // ...
        }
      },
      loading: displayLoading;
      },
    );
  }
);
```

I made some small changes:
- add delay to `fetchUser` (only in example subproject), so that it is simpler to debug `ResourceBuilder`.
- use CircularProgressIndicator when `refreshing` is true, in example and docs.

What do you think of this functionality? Also, do you know why a test I marked with `// TODO: should be errorFinder(refreshing: true)` fails with `refreshing:true` (maybe it is correct so, I am not fully sure)?

### DIFF
--- a/packages/flutter_solidart/example/lib/pages/resource.dart
+++ b/packages/flutter_solidart/example/lib/pages/resource.dart
@@ -66,18 +66,20 @@ class _ResourcePageState extends State<ResourcePage> {
             const SizedBox(height: 16),
             ResourceBuilder(
               resource: user,
-              builder: (_, userValue) {
-                return userValue.on(
-                  ready: (data, refreshing) {
+              builder: (_, userState) {
+                return userState.on(
+                  ready: (data) {
                     return Column(
                       mainAxisSize: MainAxisSize.min,
                       children: [
                         ListTile(
                           title: Text(data),
-                          subtitle: Text('refreshing: $refreshing'),
+                          subtitle:
+                              Text('refreshing: ${userState.isRefreshing}'),
                         ),
-                        if (refreshing) const CircularProgressIndicator(),
-                        if (!refreshing)
+                        if (userState.isRefreshing)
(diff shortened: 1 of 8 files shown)


manuel-plavsic:
I agree with you that using `resource.state.isRefreshing` feels like the better approach.

I implemented the changes! 🎉

Note: I had to replace `Future.microtask` with `Future.delayed(Duration(...))` and add a duration to `tester.pumpAndSettle` to make one test pass correctly. The description of the test is `"ResourceBuilder widget works properly in ResourceError state"`

packages/flutter_solidart/example/lib/pages/resource.dart
```diff
@@ -16,7 +16,13 @@ class _ResourcePageState extends State<ResourcePage> {
   @override
   void initState() {
     super.initState();
-    user = createResource(fetcher: fetchUser, source: userId);
+    user = createResource(
+      fetcher: () async {
+        await Future.delayed(const Duration(seconds: 2));
```
nank1ro, line 21: revert is as before and put the delay in the `fetchUser` function

nank1ro (approved):
Seems good to me! 🎉 I just wrote 2 request changes that might be needed

PR #36, opened 2023-06-13
[feat] hasPreviousValue getter

manuel-plavsic:
This PR adds the `hasPreviousValue` getter suggested in https://github.com/nank1ro/solidart/pull/27#issuecomment-1588990715.

It is a bit unclear to me whether or not I should make a `reportObserved` call before returning `_hasPreviousValue` (similarly to what the `previousValue` getter does).

### DIFF
--- a/packages/solidart/lib/src/core/read_signal.dart
+++ b/packages/solidart/lib/src/core/read_signal.dart
@@ -33,6 +33,10 @@ class ReadSignal<T> extends Atom implements SignalBase<T> {
 
   final T _value;
 
+  /// Whether or not there has already been a previous value. It is especially
+  /// helpful if [T] is nullable.
+  bool hasPreviousValue = false;
+
   /// All the observers
   @internal
   final List<ObserveCallback<T>> listeners = [];
--- a/packages/solidart/lib/src/core/signal.dart
+++ b/packages/solidart/lib/src/core/signal.dart
@@ -133,8 +133,6 @@ class Signal<T> extends ReadSignal<T> {
   // Tracks the internal previous value
   T? _previousValue;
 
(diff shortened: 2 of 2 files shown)


manuel-plavsic:
Done :)

nank1ro:
Thank you so much for your contributions, I really appreciate it! 👏

nank1ro (approved):

manuel-plavsic:
Btw, I think that we still need a getter, otherwise one could manually change the value of `hasPreviousValue`, for example:

```dart
  @override
  void initState() {
    super.initState();
    counter = createSignal(0);
    counter.hasPreviousValue = true;
  }
```

I will make another PR soon if it is fine for you.

That is so good to hear, thanks! I really enjoy contributing to this amazing library! Btw, I sent you a DM on Twitter a few days ago ;)

nank1ro:
You're right I hadn't thought about it that the value can be changed from outside.

Maybe you can put a getter in the `ReadSignal` that returns false.
Then in the `Signal` you may have a private `_hasPreviousSignal` value and the override the getter returning this private variable.

I didn't get any DM on Twitter 😕 , this is my profile https://twitter.com/nank1ro

manuel-plavsic:
I have a question regarding the next pull request: why should the getter in `ReadSignal` return false instead of "correctly" returning true or false. Also, I don't understand why `previousValue` is a getter that returns null. I see that it has something to do with it being a `ReadSignal` and not a `WriteSignal`, but the value held by the `ReadSignal`  can also change.

nank1ro:
Don't worry I'm going to do the PR myself cause I have to release a new dev.

PR #38, opened 2023-06-14
[chore] Rename `resource` param to `state`

manuel-plavsic:
I think that `state` is much more suitable than the current `resource`. Also, it actually is the  `resource.state` itself after all...

This will be especially helpful when using autocomplete.


manuel-plavsic:
I just thought that we could also rename it to `resourceState` or remove the param entirely and simply use `resource.state`... What do you think?

nank1ro:
I already did this change in `solidart_lint` for the __Wrap with ResourceBuilder__ assist but I forgot to propagate the change here. Name it `resourceState` like in `solidart_lint`.
`state` is not good because it conflicts with the `state` of the StatefulWidget

nank1ro (approved):

PR #43, opened 2023-06-18
[chore] Correct `Computed`'s `hasPreviousValue` getter

manuel-plavsic:
Currently, the returned value is always true, since the first time a value is set, `hasPreviousValue` is also set to true.

The PR ensures that `hasPreviousValue` returns true starting from the second time a value is — non-manually — set, i.e., the first time a value is updated.

In the first commit (fee2a4b270148b134fb281178411e0164ca1ccc8) can cause unnecessary rebuilds. The second commit (c3be4db) fixes this issue.

### DIFF
--- a/packages/solidart/lib/src/core/computed.dart
+++ b/packages/solidart/lib/src/core/computed.dart
@@ -81,6 +81,10 @@ class Computed<T> extends Signal<T> implements Derivation {
   // Whether or not there is a previous value
   bool _hasPreviousValue = false;
 
+  /// Used internally to determine if it is already possible to set
+  /// [_hasPreviousValue] to true
+  bool _computedFirstValue = false;
+
   @override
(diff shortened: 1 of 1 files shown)


nank1ro:
That's wrong.
The first value of `computed` is set here https://github.com/nank1ro/solidart/pull/43/files#diff-ff1336af3256cbb826fb12484c57514aeec782f3a3fc8c6a0f8d4a77a5572fe0L73

Please provide some code that shows the issue so I can easily understand how isn't working as expected.

manuel-plavsic:
That is right. I was a bit unsure because the initial value was always `true`. I gave an example in the other PR.

The issue has something to do with `SignalBuilder` (see https://github.com/nank1ro/solidart/pull/42#issuecomment-1598945357).

PR #45, opened 2023-06-22
[bugfix] previous value and has previous value of computed and refactor logic

nank1ro:
Refactor the core of the library in order to fix issues with `previousValue` and `hasPreviousValue` of `Computed` and simplify the logic.


nank1ro:
@manuel-plavsic

This is the only critical change I am not sure about
https://github.com/nank1ro/solidart/pull/45/files#diff-ff1336af3256cbb826fb12484c57514aeec782f3a3fc8c6a0f8d4a77a5572fe0R210

If T is something nullable I may skip the value.

nank1ro:
Yep unfortunately I just confirmed this bug with a new test:
```dart
    test('nullable derived signal', () async {
      final count = createSignal<int?>(0);
      final doubleCount = createComputed(() {
        if (count() == null) return null;
        return count()! * 2;
      });

      await pumpEventQueue();
      expect(doubleCount.value, 0);

      count.set(1);
      await pumpEventQueue();
      expect(doubleCount.value, 2);

      count.set(null);
      await pumpEventQueue();
      expect(doubleCount.value, null); // <-- FAILS HERE, IT'S 2 AND NOT NULL
    });

```

manuel-plavsic:
Yes, it is because null is supposed to be a possible value in case `T` is nullable. I think that using `T?` is more pragmatic than using the wrapper class I suggested in a previous PR, however it makes these things a bit less intuitive.

I would change

```dart
if (changed && newValue != null) {
  _setValue(newValue);
}
```

to:

```dart
if (changed) {
   _setValue(newValue as T);
}
```

I haven't run the tests yet, but I think it should work.

manuel-plavsic:
Actually, I am not so sure my suggestion works with non-nullable `T`.

nank1ro:
I actually fixed with
```dart
    if (changed && newValue is T) {
      _setValue(newValue);
    }
```

This is safe, the new test passes! thank you so much for your suggestions 💙

PR #55, opened 2023-07-06
Fix/solid signal computed

nank1ro:
- Allow getting a computed signal from a `Solid` widget
- Improve `Solid` example


manuel-plavsic:
What is the benefit of `createComputed` returning `Computed` instead of `ReadSignal`?

nank1ro:
@manuel-plavsic
Previously I had done a workaround to allow you to get a Computed with the `ReadSignal` type but then I found that this workaround did not work in release mode.
In Solid to get a provider I do a strict type comparison so `ReadSignal<T> != Computed<T>`.
I am beginning to think, however, that this is not good enough for the SolidProviders.
I should be able to get an implementation or extension of a base class

manuel-plavsic:
I understand the problematic. This happens because of the provider pattern: type safety is lost automatically. I am in favor of comparing with `is`  together with IDs when needed, but not forcing users to use IDs for every signal. E.g. if `context.get<ReadSignal<int>>() ` is called and there are multiple provided `SignalBase` instances of the same generic type (e.g. `Signal<int>` and `Computed<int>`), then IDs should be used to be sure to get the right one.

I also think that having a separate `Computed` type is beneficial, so that IDs are needed even less (since one will likely not call `context.get<ReadSignal<T>>() `; instead, they will call `context.get<Signal<T>>() ` or `context.get<Computed<T>>() `).

I agree: using `is` would be good for solid providers; sometimes we want to abstract away some implementation details, and using the interface or the super class might make more sense.

PR #73, opened 2023-11-23
Bugfix update value should not notify listeners

9dan:
Fixes a bug that cause an update of the value to notify the observers of the signal


nank1ro (approved):

nank1ro:
@all-contributors please add @9dan for code, bug, doc

PR #85, opened 2024-01-02
feat/v2.0.0

nank1ro:
- **CHORE**: Improved `Solid` widget performance by more than **3000%** in finding ancestor providers.
- **FEAT**: The `SignalBuilder` widget now automatically tracks the `Signal`s used in the `builder` function allowing you to react to N signals at the same time.
  _Before_:
  ```dart
  SignalBuilder(
    signal: counter,
    builder: (context, value, child) {
      return Text('$value');
    },
  ),
  ```

  _Now_:
  ```dart
  SignalBuilder(
    builder: (context, child) {
      return Text('${counter.value}');
    },
  ),
  ```
- **BREAKING CHANGE**: Removed `DualSignalBuilder` and `TripleSignalBuilder` in favor of `SignalBuilder`.
- **BREAKING CHANGE** the `context.observe` methods (due to the performance of the `Solid` widget) now needs the type of signal
  _Before_:
  ```dart
  final counter = context.observe<int>();
  ```
  
  _Now_:
  ```dart
  final counter = context.observe<int, Signal<int>>();
  // or
  final counter = context.observeSignal<int>();
  ```
  So this change includes the new `observeSignal`, `observeReadSignal`, `observeComputed`, `observeResource`, `observeListSignal`, `observeSetSignal` and `observeMapSignal`
  
  
  ## The PR has still some problems:
  1. The auto dispose of signals happens when using `SignalBuilder` because the effect tracks the dependencies in the `build` method. This is necessary to make it performant but, for a little moment, the signals will not be watched by the `Effect`. I still need to find a way to fix this.
  2. I don't know if `context.observe<T, S>` still be exposed, due to the new specific observe methods.
  3. Still waiting for the v2 of SolidJS that may happen at the end of January 2024, this could improve the performance of the automatic tracking system significantly (https://github.com/bubblegroup/bubble-reactivity)

### DIFF
--- a/examples/counter/lib/main.dart
+++ b/examples/counter/lib/main.dart
@@ -41,9 +41,8 @@ class _CounterPageState extends State<CounterPage> {
       ),
       body: Center(
         child: SignalBuilder(
-          signal: counter,
-          builder: (_, value, __) {
-            return Text('$value');
+          builder: (_, __) {
+            return Text(counter().toString());
           },
         ),
       ),
--- a/examples/counter/pubspec.yaml
+++ b/examples/counter/pubspec.yaml
@@ -31,7 +31,7 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
-  flutter_solidart: ^1.7.0
+  flutter_solidart: ^2.0.0
 
   # The following adds the Cupertino Icons font to your application.
   # Use with the CupertinoIcons class for iOS style icons.
--- a/examples/github_search/pubspec.yaml
+++ b/examples/github_search/pubspec.yaml
@@ -30,7 +30,7 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
-  flutter_solidart: ^1.7.0
+  flutter_solidart: ^2.0.0
   json_annotation: ^4.8.1
   equatable: ^2.0.5
   http: ^1.0.0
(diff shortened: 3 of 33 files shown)


nank1ro:
> Note that `SignalBuilder` has a hidden footgun that will catch some users. Any signals accessed in its entire subtree will trigger a rebuild in the parent.
> 
> 
> 
> ```fsharp
> 
> SignalBuilder
> 
>   |_ Container
> 
>       |_ Column
> 
>             |_ SizedBox
> 
>                  |_ SignalBuilder
> 
>                        |_ Text  <- signal watched here will rebuild both SignalBuilders
> 
> ``` 

Thanks for reporting 🙏, need to investigate further to fix this

nank1ro:
@jinyus can't reproduce the issue, this is the code I used to test
```dart
import 'package:flutter/material.dart';
import 'package:flutter_solidart/flutter_solidart.dart';

final a = Signal(0);
final b = Signal(100);

class TestPage extends StatelessWidget {
  const TestPage({super.key});

  @override
    return Scaffold(
      body: SignalBuilder(
        builder: (_, __) {
          print('build a');
          return Column(
            children: [
              Text(a().toString()),
              const Second(),
            ],
          );
        },
      ),
      floatingActionButton: Column(
        mainAxisSize: MainAxisSize.min,
        children: [
          TextButton(
            onPressed: () => a.value++,
            child: const Text('add to A'),
          ),
          TextButton(
            onPressed: () => b.value++,
            child: const Text('add to B'),
          ),
        ],
      ),
    );
  }
}

class Second extends StatelessWidget {
  const Second({super.key});

  @override
    return SignalBuilder(
      builder: (_, __) {
        print('build b');
        return Text(b().toString());
      },
    );
  }
}
```
Can you share your code please?

jinyus:
I'm no longer seeing this bug. I will have to investigate further. Consider this a non-issue for now.

PR #97, opened 2024-12-28
[Chore] Create separate stream constructor for Resource

manuel-plavsic:
This PR makes it simpler to work with `Resource`: the users are forced to provide the `fetcher` with the default constructor (like with SolidJS). In case the user wants to provide a stream, then a different constructor is needed. This way it becomes impossible to provide:

- both `stream` and `fetcher`
- none of them

By getting rid of the two scenarios above, type inference also becomes very accurate.

This PR introduced a breaking change, but I think it is fine, since v2 will bring other breaking changes as well.


nank1ro (approved):
LGTM! Thanks for the PR 🙏

PR #100, opened 2024-12-30
[Chore] Rename Provider-related classes

manuel-plavsic:
I think the name `Solid` for the widget is a bit general/unclear. We should use something more clear like `ProviderScope`. The same concept holds for `SolidElement`, which is now `ProviderElement`. I still need to change them throughout the code base and the documentation, and rename the files. But what do you think? If you like my suggestions then I'll happily adapt them everywhere :)

Do you have a better name for `ProviderScope.value`? Maybe `ProviderScope.separate`, which hints at a separate tree?

I added some constructors to it to make them more type safe (a bit like the other PR that just got merged).

### DIFF
--- a/packages/flutter_solidart/lib/src/widgets/solid.dart
+++ b/packages/flutter_solidart/lib/src/widgets/solid.dart
@@ -6,7 +6,7 @@ import 'package:flutter_solidart/flutter_solidart.dart';
 
 part '../models/solid_element.dart';
 
-/// {@template solid}
+/// {@template provider-scope}
 /// Provides [providers] to descendants.
 ///
 /// The Flutter framework works like a Tree. There are ancestors and there are
@@ -34,7 +34,7 @@ part '../models/solid_element.dart';
 ///   @override
 ///   Widget build(BuildContext context) {
 ///     // Provide the theme mode signal to descendats
-///     return Solid(
+///     return ProviderScope(
 ///       providers: [
 ///         Provider<Signal<ThemeMode>>(
 ///           create: () => Signal(ThemeMode.light),
@@ -150,7 +150,7 @@ part '../models/solid_element.dart';
 ///       appBar: AppBar(
 ///         title: const Text('Providers'),
 ///       ),
-///       body: Solid(
+///       body: ProviderScope(
 ///         providers: [
 ///           Provider<NameProvider>(
 ///             create: () => const NameProvider('Ale'),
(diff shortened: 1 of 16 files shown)


packages/flutter_solidart/lib/src/widgets/solid.dart
```diff
@@ -246,21 +246,25 @@ part '../models/solid_element.dart';
...
     super.key,
     this.child,
+    this.providers = const [],
+  })  : builder = null,
+        _canAutoDisposeProviders = true;
+
+  /// {@macro provider-scope}
+  const ProviderScope.builder({
+    super.key,
     this.builder,
```
nank1ro, line 261: I don't like the idea to separate the builder from the main constructor.
At the end, they are just two different methods to pass a child.
If you use the builder function you get immediately a context which can see the provided providers, with child not.

packages/flutter_solidart/lib/src/widgets/solid.dart
```diff
@@ -272,21 +276,21 @@ class Solid extends StatefulWidget {
         ),
         _canAutoDisposeProviders = autoDispose;
 
-  /// Provide a single or multiple [SolidElement]s to a new route.
+  /// Provide a single or multiple [ProviderElement]s to a new route.
   ///
   /// This is useful for passing multiple providers to modals, because are
   /// spawned in a new tree.
-  factory Solid.value({
+  factory ProviderScope.value({
```
nank1ro, line 283: ProviderScope.value is not a bad name. It's similar to `Provider.value`, the `provider` package.

nank1ro (commented):
Of course ProviderScope is more clear and for who comes from riverpod it's straightforward. The only downside I can think of is that it's more verbose than just `Solid`

manuel-plavsic:
> Of course ProviderScope is more clear and for who comes from riverpod it's straightforward. The only downside I can think of is that it's more verbose than just `Solid`

I think `Solid` is really not a great choice, especially to new Flutter users. `ProviderScope` is a bit longer, but it doesn't feel too verbose IMO.

packages/flutter_solidart/lib/src/widgets/solid.dart
manuel-plavsic: I didn't know this. Let's keep it then.

manuel-plavsic:
Closing this in favor of https://github.com/nank1ro/solidart/pull/101, which includes these changes and more.

PR #101, opened 2024-12-30
[Feature] provider context

manuel-plavsic:
This PR introduces ProviderContexts, which are IDs with first class support. It allows providers to be typesafe. They are similar to createContext/useContext in Solid.js.

They can still throw if absent while injecting them, but at least the creation and the injection can be done in a typesafe way.

More examples and tests are needed. I also did not test them manually yet, but this is a first draft and I was excited to show it.

NB: it might be easiest to look at the differences by looking at the last commit. While I was working on this, I was also renaming Provider-related classes, constructors, ...

```dart
  // the context (the id is global
  const myCounter = ProviderContext<Signal<int>>();

 // add it with a ProviderScope
 ProviderScope(
   providers: [ Provider.context(myCounter, create: ...) ],
   child: ...
 )
   
  // some function inside some widget
  someFn(BuildContext context) {
    context.updateByCtx(myCounter, (value) => value * 2);
  }
```

This is equal to:

```dart
// retrieve the signal
final signal = context.getByCtx(myCounter);
// update the signal
signal.update((value) => value * 2);
```

### DIFF
--- a/examples/github_search/lib/ui/search_page.dart
+++ b/examples/github_search/lib/ui/search_page.dart
@@ -9,7 +9,7 @@ class SearchPage extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    return Solid(
+    return ProviderScope(
       providers: [
         // Provide the [GithubSearchBloc] to descendants
         Provider<GithubSearchBloc>(
--- a/examples/todos/lib/todos_page.dart
+++ b/examples/todos/lib/todos_page.dart
@@ -10,7 +10,7 @@ class TodosPage extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     // Using Provider here to provide the [TodosController] to descendants.
-    return Solid(
+    return ProviderScope(
       providers: [
         Provider<TodosController>(
           create: () => TodosController(initialTodos: Todo.sample),
--- a/examples/todos/lib/widgets/todos_body.dart
+++ b/examples/todos/lib/widgets/todos_body.dart
@@ -27,7 +27,7 @@ class _TodosBodyState extends State<TodosBody> {
     // in both the `initState` and `build` methods.
     final todosController = context.get<TodosController>();
 
-    return Solid(
+    return ProviderScope(
       providers: [
         // make the active filter signal visible only to descendants.
         // created here because this is where it starts to be necessary.
--- a/examples/todos/test/widget_test.dart
+++ b/examples/todos/test/widget_test.dart
@@ -20,7 +20,7 @@ Widget wrapWithMockedTodosController({
   required TodosController todosController,
 }) {
   return MaterialApp(
-    home: Solid(
+    home: ProviderScope(
       providers: [
         Provider<TodosController>(
           create: () => todosController,
--- a/examples/toggle_theme/lib/main.dart
+++ b/examples/toggle_theme/lib/main.dart
@@ -11,7 +11,7 @@ class MyApp extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     // Provide the theme mode signal to descendats
-    return Solid(
+    return ProviderScope.builder(
       providers: [
         Provider<Signal<ThemeMode>>(
           create: () => Signal(ThemeMode.light),
--- a/packages/flutter_solidart/example/lib/pages/solid/multiple_signals_page.dart
+++ b/packages/flutter_solidart/example/lib/pages/solid/multiple_signals_page.dart
@@ -11,7 +11,7 @@ class MultipleSignalsPage extends StatelessWidget {
       appBar: AppBar(
         title: const Text('Multiple Signals'),
       ),
-      body: Solid(
+      body: ProviderScope(
         providers: [
           // provide the firstName signal to descendants
           Provider<Signal<String>>(
--- a/packages/flutter_solidart/example/lib/pages/solid/observe_signal_page.dart
+++ b/packages/flutter_solidart/example/lib/pages/solid/observe_signal_page.dart
@@ -10,7 +10,7 @@ class ObserveSignalPage extends StatelessWidget {
       appBar: AppBar(
         title: const Text('Observe Signal'),
       ),
-      body: Solid(
+      body: ProviderScope(
         providers: [
           // provide the count signal to descendants
           Provider<Signal<int>>(create: () => Signal(0)),
--- a/packages/flutter_solidart/example/lib/pages/solid/solid_providers.dart
+++ b/packages/flutter_solidart/example/lib/pages/solid/solid_providers.dart
@@ -26,7 +26,7 @@ class ProvidersPage extends StatelessWidget {
       appBar: AppBar(
         title: const Text('Providers'),
       ),
-      body: Solid(
+      body: ProviderScope(
         providers: [
           Provider<NameProvider>(
             create: () => const NameProvider('Ale'),
@@ -58,7 +58,7 @@ class SomeChild extends StatelessWidget {
   Future<void> openDialog(BuildContext context) {
     return showDialog(
       context: context,
-      builder: (_) => Solid.value(
+      builder: (_) => ProviderScope.value(
         elements: [
           context.getElement<NameProvider>(),
           context.getElement<NumberProvider>(#firstNumber),
--- a/packages/flutter_solidart/example/lib/pages/solid/solid_reactivity.dart
+++ b/packages/flutter_solidart/example/lib/pages/solid/solid_reactivity.dart
@@ -13,7 +13,7 @@ class SolidReactivityPage extends StatefulWidget {
 class _SolidReactivityPageState extends State<SolidReactivityPage> {
   @override
   Widget build(BuildContext context) {
-    return Solid(
+    return ProviderScope(
       providers: [
         Provider<Signal<int>>(create: () => Signal(0), id: #firstCounter),
         Provider<Signal<int>>(create: () => Signal(0), id: #secondCounter),
--- a/packages/flutter_solidart/example/lib/pages/solid/solid_signals.dart
+++ b/packages/flutter_solidart/example/lib/pages/solid/solid_signals.dart
@@ -23,7 +23,7 @@ class _SolidSignalsPageState extends State<SolidSignalsPage> {
       appBar: AppBar(
         title: const Text('Solid Signals'),
       ),
-      body: Solid(
+      body: ProviderScope(
         providers: [
           // provide the count signal to descendants
           Provider<Signal<int>>(
--- a/packages/flutter_solidart/lib/flutter_solidart.dart
+++ b/packages/flutter_solidart/lib/flutter_solidart.dart
@@ -3,6 +3,7 @@ library flutter_solidart;
 
 export 'package:solidart/solidart.dart';
 
+export 'src/models/provider_context.dart';
 export 'src/utils/extensions.dart';
 export 'src/utils/solid_extensions.dart';
 export 'src/widgets/show.dart';
--- a/packages/flutter_solidart/lib/src/models/provider_context.dart
+++ b/packages/flutter_solidart/lib/src/models/provider_context.dart
@@ -1,0 +1,15 @@
+import 'package:flutter_solidart/flutter_solidart.dart';
+
+/// {@template provider-context}
+/// A [ProviderContext] is a type-safe identifier for [Provider]s.
+///
+/// To use [ProviderExtensions.observeByCtx] or [ProviderExtensions.updateByCtx]
+/// with this provider context, the type [T] must be a [SignalBase].
+///
+/// A [ProviderContext] is the equivalent of a
+/// [Context in SolidJS](https://docs.solidjs.com/concepts/context).
+/// {@endtemplate}
+class ProviderContext<T> {
+  /// {@macro provider-context}
+  const ProviderContext();
+}
--- a/packages/flutter_solidart/lib/src/models/solid_element.dart
+++ b/packages/flutter_solidart/lib/src/models/solid_element.dart
@@ -9,12 +9,12 @@ typedef DisposeValue<T> = void Function(T value);
 /// The idenfifier of the element.
 typedef Identifier = Object;
 
-/// {@template solidelement}
+/// {@template provider-element}
 /// The base class of a solid provider
 /// {@endtemplate}
-abstract class SolidElement<T> {
-  /// {@macro solidelement}
-  const SolidElement({
+abstract class ProviderElement<T> {
+  /// {@macro provider-element}
+  const ProviderElement({
     required this.create,
     this.id,
   });
@@ -29,7 +29,7 @@ abstract class SolidElement<T> {
   /// Returns the type of the value
   Type get _valueType => T;
 
-  bool get _isSignal => this is SolidElement<SignalBase>;
+  bool get _isSignal => this is ProviderElement<SignalBase>;
 
   void _disposeFn(BuildContext context, dynamic value);
 }
@@ -42,7 +42,7 @@ typedef SolidProvider<T> = Provider<T>;
 
 /// {@template provider}
 /// A Provider that manages the lifecycle of the value it provides by
-// delegating to a pair of [create] and [dispose].
+/// delegating to a pair of [create] and [dispose].
 ///
 /// It is usually used to avoid making a StatefulWidget for something trivial,
 /// such as instantiating a BLoC.
@@ -61,7 +61,7 @@ typedef SolidProvider<T> = Provider<T>;
 ///
 /// {@endtemplate}
 @immutable
-class Provider<T> extends SolidElement<T> {
+class Provider<T> extends ProviderElement<T> {
   /// {@macro provider}
   const Provider({
     required super.create,
--- a/packages/flutter_solidart/lib/src/models/solid_element.dart
+++ b/packages/flutter_solidart/lib/src/models/solid_element.dart
@@ -70,6 +70,14 @@ class Provider<T> extends ProviderElement<T> {
     super.id,
   });
 
+  /// {@macro provider}
+  const Provider.context(
+    ProviderContext<T> context, {
+    required super.create,
+    this.dispose,
+    this.lazy = true,
+  }) : super(id: context);
+
   /// An optional dispose function called when the Solid that created this
   /// provider disposes
   final DisposeValue<T>? dispose;
--- a/packages/flutter_solidart/lib/src/utils/solid_extensions.dart
+++ b/packages/flutter_solidart/lib/src/utils/solid_extensions.dart
@@ -1,30 +1,30 @@
 import 'package:flutter/material.dart';
 import 'package:flutter_solidart/flutter_solidart.dart';
 
-/// Convenience extensions to interact with the [Solid] InheritedModel.
-extension SolidExtensions on BuildContext {
-  /// {@macro solid.get}
+/// Convenience extensions to interact with the [ProviderScope] InheritedModel.
+extension ProviderExtensions on BuildContext {
+  /// {@macro provider-scope.get}
   P get<P>([Identifier? id]) {
-    return Solid.get<P>(this, id);
+    return ProviderScope.get<P>(this, id);
   }
 
-  /// {@macro solid.maybeGet}
+  /// {@macro provider-scope.maybeGet}
   P? maybeGet<P>([Identifier? id]) {
-    return Solid.maybeGet<P>(this, id);
+    return ProviderScope.maybeGet<P>(this, id);
   }
 
-  /// {@macro solid.getElement}
-  SolidElement<P> getElement<P>([Identifier? id]) {
-    return Solid.getElement<P>(this, id);
+  /// {@macro provider-scope.getElement}
+  ProviderElement<P> getElement<P>([Identifier? id]) {
+    return ProviderScope.getElement<P>(this, id);
   }
 
-  /// {@macro solid.observe}
+  /// {@macro provider-scope.observe}
   T observe<T extends SignalBase<dynamic>>([Identifier? id]) {
-    return Solid.observe<T>(this, id);
+    return ProviderScope.observe<T>(this, id);
   }
 
-  /// {@macro solid.update}
+  /// {@macro provider-scope.update}
   void update<T>(T Function(T value) callback, [Identifier? id]) {
-    return Solid.update<T>(this, callback, id);
+    return ProviderScope.update<T>(this, callback, id);
   }
 }
--- a/packages/flutter_solidart/lib/src/utils/solid_extensions.dart
+++ b/packages/flutter_solidart/lib/src/utils/solid_extensions.dart
@@ -8,23 +8,67 @@ extension ProviderExtensions on BuildContext {
     return ProviderScope.get<P>(this, id);
   }
 
+  /// {@macro provider-scope.get}
+  P getByCtx<P>(ProviderContext<P> providerContext) => get<P>(providerContext);
+
   /// {@macro provider-scope.maybeGet}
   P? maybeGet<P>([Identifier? id]) {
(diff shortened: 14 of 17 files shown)


nank1ro (commented):
Thanks for the detailed explanation and the PR 🙏
Don't you think there will be to many ways to provide an instance to the subtree in this way?

We may remove the other standard way in favor of this, but I would prefer to avoid adding additional complexity to the user.
One way to do a thing is better than two or three at the same time.

If this solution is more compile safe, then we should throw all the other code in favor of this one only.

manuel-plavsic:
> Thanks for the detailed explanation and the PR 🙏 Don't you think there will be to many ways to provide an instance to the subtree in this way?
> 
> We may remove the other standard way in favor of this, but I would prefer to avoid adding additional complexity to the user. One way to do a thing is better than two or tree at the same time.
> 
> If this solution is more compile safe, then we should throw all the other code in favor of this one only.

It was my pleasure :) If we don't remove the other way, there will be many ways, yes. I like your suggestion: let's have only one implementation (which is definitely less confusing, I agree). Yes, this one is more compile safe. I will replace the previous implementation with this one (so we'll have only e.g. `context.get<T>(ProviderContext<T> providerContext)` in the end).  I'll also add examples and tests.

manuel-plavsic:
Since we are removing the other implementation, there is no longer need for a non-typesafe `Identifier? id`. In this case, I think we can rename `ProviderContext` to `ProviderId`. It would no longer be like in Solid.js, but maybe it is more clear this way (and also shorter). Which one do you prefer?

nank1ro:
> Since we are removing the other implementation, there is no longer need for a non-typesafe `Identifier? id`. In this case, I think we can rename `ProviderContext` to `ProviderId`. It would no longer be like in Solid.js, but maybe it is more clear this way (and also shorter). Which one do you prefer?

I don't like ProviderContext because the BuildContext is another thing and may confuse. At the end it's just an identifier

nank1ro:
It's not clear how do you get the inferred type correctly when using `final counter = context.get(myCounter)`

Do you still need to specify T?
Otherwise I don't understand how counter knows its type

manuel-plavsic:
No need to specify the type every time. It is done only once, when the global identifier is defined (e.g. `const myCounter = ProviderContext<Signal<int>>();`). Note that the identifier `ProviderId` takes a generic type. If `ProviderId` didn't have the generic type, then the type inference would not be possible.

Since `ProviderId` has the generic type, when you use `context.get<P>(ProviderId<P> providerContext)`, the code in the PR description example infers the type to `context.get<Signal<int>>`, since `myCounter` is an instance of `ProviderId<Signal<int>>`. And since `context.get<T>` returns a `T`, it returns a `Signal<int>>` in the `myCounter` case.

I think this will be extremely convenient when dealing with long or many nested generic types.

examples/github_search/lib/ui/search_page.dart
```diff
@@ -4,16 +4,18 @@ import 'package:github_search/bloc/github_search_bloc.dart';
...
   @override
   Widget build(BuildContext context) {
-    return Solid(
+    return ProviderScope(
       providers: [
         // Provide the [GithubSearchBloc] to descendants
-        Provider<GithubSearchBloc>(
-          create: () => GithubSearchBloc(),
+        _githubSearchBlocId.createProvider(
+          init: () => GithubSearchBloc(),
```
nank1ro, line 18: Could we have
```
Provider(githubSearchBlocId, create: () {}, ...)
```
?

examples/todos/lib/controllers/controller.dart
```diff
@@ -14,6 +14,8 @@ class TodosController {
     List<Todo> initialTodos = const [],
   }) : todos = ListSignal(initialTodos);
 
+  static final ProviderId<TodosController> id = ProviderId();
```
nank1ro, line 17: I like it!

`static final id = ProviderId<TodosController>()` it's pretty nice.

The usage would be `TodosController.id`

examples/github_search/lib/ui/search_page.dart
nank1ro, line 18: create is a convention introduced by Provider (the package). Even copilot is used to type it. I would keep it as it is

manuel-plavsic: It is possible, but `Provider(githubSearchBlocId, create: () {}, ...)` will result in a `Provider<dynamic>(githubSearchBlocId, create: () {}, ...)` if passed directly as argument without creating a variable for it (probably 99% of cases). Which means that either:

- you will need to add the type e.g. `Provider<GithubSearchBlocId>(githubSearchBlocId, create: () {}, ...)`, which becomes very long.
- you don't add the type, but now the create function can return anything, e.g. an `int`. The error will be noticed only at runtime.

I renamed `create` to `init` because the provider gets created immediately, but it is initialized later if it is lazy. I know it is a breaking change, but I think it would make sense to correct it.

manuel-plavsic: If you want to, we can still add a the constructor with the `create` param, and we can deprecate it.

nank1ro: > I renamed create to init because the provider gets created immediately, but it is initialized later if it is lazy

What do you mean with "the provider gets created immediately"? I haven't noticed nothing different from before

manuel-plavsic: I see. But in case we keep `id.createProvider`, I think that having `create` immediately again could be a bit ugly

nank1ro: Can you provide an example for your statement "if passed directly as argument without creating a variable for it (probably 99% of cases)"?

examples/github_search/lib/ui/search_page.dart
```diff
@@ -55,7 +57,7 @@ class _SearchBar extends StatefulWidget {
 class __SearchBarState extends State<_SearchBar> {
   final textController = TextEditingController();
   // retrieve the ancestor [GithubSearchBloc]
-  late final bloc = context.get<GithubSearchBloc>();
+  late final bloc = _githubSearchBlocId.get(context);
```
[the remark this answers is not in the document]
manuel-plavsic: It can be done. I have to say thought that when autocomplete, as soon as you choose e.g. `get` it gets autocompleted nicely ( `_githubSearchBlocId.g` -> `_githubSearchBlocId.get(context)`).

Would you support both? Only `context.get(githubSearchBlocId)`? Or both with `context.get(githubSearchBlocId)` deprecated?

examples/todos/lib/controllers/controller.dart
manuel-plavsic: Glad you like it :) Yes, exactly

examples/github_search/lib/ui/search_page.dart
manuel-plavsic: > > I renamed create to init because the provider gets created immediately, but it is initialized later if it is lazy
> 
> What do you mean with "the provider gets created immediately"? I haven't noticed nothing different from before

I thought about the instance of `Provider`, not what is actually going on under the hood with `HashMap`. I guess we can call it `create` then, yes.

examples/github_search/lib/ui/search_page.dart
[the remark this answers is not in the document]
nank1ro: No just one way. Two ways would be very strange and sounds like something is something wrong. What about people who review PRs having the mixed usage of both? A nightmare

examples/github_search/lib/ui/search_page.dart
manuel-plavsic: > Can you provide an example for your statement "if passed directly as argument without creating a variable for it (probably 99% of cases)"?

I assume most people will likely write

```
ProviderScope.builder(
  providers: [
    nameProviderId.createProvider(
      create: () => "name",
    ),
  ],
  // ...
)
```

over

```
// this is not necessarily global, just somewhere above the ProviderScope
final nameProvider = nameProviderId.createProvider(
      create: () => "name",
);

// ...

ProviderScope.builder(
  providers: [
    nameProvider,
  ],
  // ...
)
```

But this is just my assumption.

manuel-plavsic: BTW, I honestly think that

```
final counter = counterId.get(innerContext);
final doubleCounter = doubleCounterId.get(innerContext);
```

is much more readable than:

```
final counter = innerContext.get(counterId);
final doubleCounter = innerContext.get(doubleCounterId);
```

since one can easily notice if the name of the variable and the one of the providerId are different. It is a potential sign of a mistake.

examples/github_search/lib/ui/search_page.dart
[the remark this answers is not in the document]
manuel-plavsic: BTW, I honestly think that

```
final counter = counterId.get(innerContext);
final doubleCounter = doubleCounterId.get(innerContext);
```

is much more readable than:

```
final counter = innerContext.get(counterId);
final doubleCounter = innerContext.get(doubleCounterId);
```

since one can easily notice if the name of the variable and the one of the providerId are different: in that case it is a potential sign of a mistake.

examples/github_search/lib/ui/search_page.dart
nank1ro: I am very conflicted about that. A little bit you are convincing me.
I was going more to context.get because it is more similar to how it is now in the v1.

Thinking about the uses of Flutter, do you have references where they do the opposite as in this proposal? Even for routers I seem to have always done context.router

Probably GoRouter as something similar where it does `MyRoute().go(context)`

examples/github_search/lib/ui/search_page.dart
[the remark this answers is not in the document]
nank1ro: I am very conflicted about that. A little bit you are convincing me.
I was going more to context.get because it is more similar to how it is now in the v1.

Thinking about the uses of Flutter, do you have references where they do the opposite as in this proposal? Even for routers I seem to have always done context.router

Probably GoRouter as something similar where it does `MyRoute().go(context)`

examples/github_search/lib/ui/search_page.dart
nank1ro: Isn't the first example impossible?
providerId has a type T.
The create should return the type T. Assuming providerId is of type `bool` and you return `String` from the create method, the analyser doesn't blame about the type?

examples/github_search/lib/ui/search_page.dart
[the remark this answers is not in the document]
manuel-plavsic: The `.of(context)` is pretty common. TypeSafe GoRouter routes are exactly like you wrote https://pub.dev/documentation/go_router/latest/topics/Type-safe%20routes-topic.html.

Sometimes, `counterId.get(innerContext)` might get inferred as `dynamic`.

Also, I work with inlay hints a lot, and `counterId.get(innerContext)` makes the code harder to read.

Screenshot From 2025-01-01 21-29-33
Screenshot From 2025-01-01 21-32-11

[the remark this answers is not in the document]
nank1ro: > Also, I work with inlay hints a lot, and counterId.get(innerContext) makes the code harder to read.

Did you make a typo right? You meant the opposite?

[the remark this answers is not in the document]
manuel-plavsic: yes, excuse me, I meant the opposite

examples/github_search/lib/ui/search_page.dart
manuel-plavsic: Why would it be impossible? If `nameProviderId` is of type `Provider<String>>`, then both examples should be valid.

manuel-plavsic: Shortly I'll add some screenshots regarding the issue with `Provider()`.

examples/github_search/lib/ui/search_page.dart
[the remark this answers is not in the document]
nank1ro: You convinced me about this ✌️
Let's go for `providerId.get(context)`

examples/github_search/lib/ui/search_page.dart
nank1ro: > Why would it be impossible? If `nameProviderId` is of type `Provider<String>>`, then both examples should be valid.

With impossible I meant the analyser shows an error because there is a type mismatch. For this reason it should not compile

manuel-plavsic: If one uses `id.createProvider`, only the right return type can be returned. If one tries something else it will be highlighted with red.
Screenshot From 2025-01-01 22-55-04

With `Provider()`, it will infer a Provider<dynamic> be default if placed directly inside the providers list (however, if you first assigned this provider to a variable it would have been fine).
Screenshot From 2025-01-01 22-56-07

And this is what happens if you leave `Provider<dynamic>` and run it:
Screenshot From 2025-01-01 22-57-10

I think this might be confusing for people that don't know the library well.

What do you think we should do?

examples/github_search/lib/ui/search_page.dart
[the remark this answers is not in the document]
manuel-plavsic: I appreciate your open-mindedness :)

examples/github_search/lib/ui/search_page.dart
manuel-plavsic: we should set

```
analyzer:
  errors:
    inference_failure_on_instance_creation: error
```

this also ensures the user specifies the generic type for `ProviderId` (we should probably add this to the docs).

I am still not so sure regarding the `Provider()` constructor. Even if one sets `inference_failure_on_instance_creation`, they still have to write the type, not only the variable.

nank1ro: I would like to play around a bit with the code tomorrow to check any of my doubts.

For now I would just make another change:
- make the init a positional argument, so we'd have `nameId.createProvider(() => 'name', lazy: false, ...)`
In this way we could also have:
`klassId.createProvider(Klass.new)`

manuel-plavsic: I think another good reason to go for the `providerId.createProvider` approach might be that it is slightly more straightforward to check if they were created. For example, if you look at the next picture:

Screenshot From 2025-01-02 11-52-43

in the second case there is that `Provider<GithubSearchBloc>(` a bit in the way, since now with the ProviderId feature, I no longer really care about the Provider type, only about if the provider was created and thus can be accessed (and I refer to the ProviderId to access it). And if there is a list of providers to analyze, the first approach is a bit easier to follow.

Also, I think if you create a variable for a provider like below, it can be argued that is a bit easier to spot mistakes since the variable name and the id are places right to each other.

Screenshot From 2025-01-02 11-57-01

This is especially noticeable with inlay hints.
 
Screenshot From 2025-01-02 11-57-23
 
I would love to hear your feedback after you play around with the code.

nank1ro: I'm working on the new reactivity system right now, I haven't tried it yet.
Watching again the code you provided makes me wondering if ProviderId is really needed.
Can't we just have

```dart
final githubSearchBlocProvider = Provider(
  create: GithubSearchBloc.new,
  dispose: (bloc) => bloc.dispose(),
);
```

In this way we could eliminate the new `ProviderId` model and keep stuff more simple, what do you think?

To get the provider I could write
```dart
final bloc = githubSearchBlocProvider.get(context);
```

So all the extensions are moved to the Provider type.

nank1ro: I know you said about the type inference, but are you sure there isn't another way to make it working correctly?
```dart
final githubSearchBlocProvider = Provider<GithubSearchBloc>(
  create: GithubSearchBloc.new,
  dispose: (bloc) => bloc.dispose(),
);
```

If I type it here, I think it can be the only place in the app where the type is needed

manuel-plavsic: I am not in favour of that.

That gives one of these problems:

- if there are no ids at all, then it won't be possible to have different providers of the same type. You could use custom wrapper types to distinguish them, but I don't like that approach. I think it is much better to work with ProviderId.
- if you include separate IDs, then cannot be typesafe. Then there will basically be no difference than how it was before.
  - theoretically you could add something like this if you want to have typesafe ids to distinguish between other providers:
    ```dart
    class ProviderHelper<T> {
      final Provider<T> provider;
      final T id;
    }
    ```
    But it is ugly IMO, and it complicates the API a lot.

Also, SolidJS introduced contexts to prevent these scenarios.

manuel-plavsic: Hold on a second. Actually, what you suggest could also work, since every provider instance would have a different hash, also if they are of the same generic type (as long as they are not instantiated with `const`). If you define them like that, you will have to make them global so that Widget can access it. However, what if you want to initialize them with a specific value that you get from a resource for example?

manuel-plavsic: I guess we could use two constructors:

```
final githubSearchBlocProvider = Provider(
  create: GithubSearchBloc.new, // same as: () => GithubSearchBloc()
  dispose: (bloc) => bloc.dispose(),
);
```

And something like this (maybe a better name):

```
final githubSearchBlocProvider = Provider.withValue(
  create: (block) => block, // this is the identity function, but there could also be a transformation here
  dispose: (bloc) => bloc.dispose(),
);
```

manuel-plavsic: I just realized there would be a potential issue with your suggestion and also with mine.

If you need more providers that work exactly the same, you can just generate more IDs.

- with your suggestion, this is just not possible. You would have to duplicate the Provider code every time you need it.
- with my suggestion, it is also not possible because the create/init, dispose, ... are in the current Provider class and not in ID.

I suggest we rename `ProviderId` to `Provider` and a `ValueProvider` (the latter extends the former), and move there the create, dispose, ... members to `Provider`.

```
final githubSearchBlocProvider = Provider(
  create: GithubSearchBloc.new, // same as: () => GithubSearchBloc()
  dispose: (bloc) => bloc.dispose(),
);
```

```
final githubSearchBlocValueProvider = ValueProvider(
  create: (block) => block, // this is the identity function, but there could also be a transformation here
  dispose: (bloc) => bloc.dispose(),
);
```

Then we offer ways to generate a `ProviderInstance` (our current `Provider`). If you have a better name suggestion please share it!

```
ProviderScope(
  providers:
    [
      githubSearchBlocProvider.newInstance(),
      githubSearchBlocValueProvider.newInstance(GithubSearchBloc(someSpecialArgs)),
    ],
  //...
)

Comment (manuel-plavsic) [reply examples/github_search/lib/ui/search_page.dart:18]: I think that in the case of ValueProvider, the `create` is not necessary:

```
final githubSearchBlocValueProvider = ValueProvider(
  dispose: (bloc) => bloc.dispose(),
);
```

And then we do:

```
githubSearchBlocValueProvider.newInstance(GithubSearchBloc(someSpecialArgs))
```

Comment (manuel-plavsic) [reply examples/github_search/lib/ui/search_page.dart:18]: I realized that what you are suggesting is actually much better than my last ideas. It is simple, beautiful and powerful. I am quickly gonna refactor then you can try the code. I'll do it exactly like suggested :)

Comment (manuel-plavsic) [reply examples/github_search/lib/ui/search_page.dart:18]: I also have an idea for the `family` alternative

Comment (nank1ro) [reply examples/github_search/lib/ui/search_page.dart:18]: Yeah I'm thinking about family but not in the same way as riverpod, let me show you some pseudo code.
Let's say that we have a TodoBloc that handles a single todo item. To work with that item (e.g. retrieve it from the server) it needs a `todoId`

```dart
TodoBloc(this.todoId);
```

The provider would be

```dart
final todoBlocProvider = Provider<GithubSearchBloc>(
  create: (todoId) => TodoBloc(todoId),
  dispose: (bloc) => bloc.dispose(),
);
```

When passed to the ProviderScope widget this would be

```dart
ProviderScope.builder(
  providers: [
    todoBlocProvider.withArguments(todoId: "todo_1",
  ],
```

I'm not sure this is feasible but I remember something like this on mobx, will add a link as soon as I find it

manuel-plavsic: I pushed the code.

I added:

- `ProviderWithArg<A, T>` (`A` is the argument type)
  - This is a `Provider` with a `setInitialArg` function. This should be used the same way as a `Provider`. I added a test for this class.

- `ProviderMap<K, T>` (`K` is the key type)
  - Map of normal `Provider`s (here the generics are first key type and then provider type, since it makes way more sense IMO).
  
- `ProviderMapWithArg<K, A, T>`
  - Map of `ProviderWithArgs`
  
- `ProviderMapWithKeyArg<KA, T>` (`KA` is both the key type as well as the argument type)
  - This is basically the `family`.
  - It is the same as `ProviderMapWithArg<KA, KA, T>`, but this way it is a bit simpler to read the generic types.

manuel-plavsic: Since macros are coming but still not here, and I'd much rather avoid `build_runner`, I think the best approach would be to just pass an argument for the moment. Then that are can be defined to be a tuple. It should be fine for the moment.

I think in the future we can do something like you suggested, when macros are in stable.

manuel-plavsic: I just assumed mobx used build runner, but now I looked a bit closer.

The mobx approach seems to use map values of `dynamic` type, so that is not really optimal. I will be very busy for the next days. Maybe you can play around with the current code and see if you manage to implement what you suggested in a type-safe manner.

nank1ro: I don't know if it will work but I'm pretty sure I can't make it type safe.
Seems like we are going into something that overcomplicates the Solid widget a lot, just for making it "compile-safe" when in reality it's not compile safe neither in our proposals. If the provider is not attached to a Solid widget, when using `provider.get(context` it would throw anyway because it cannot find it.

I would put this PR on hold until we have a **simple** proposal that really brings benefits.
The addition of family providers makes it much more complicated for the user, who must necessarily be familiar with all the different types of providers, I don't like it. 

I hope you feel the same way.

I will quote the purposes of the library, which must remain so:
> The objectives of this library are:
Being simple and easy to learn
Fits well with the framework's good practices
Do not have a single global state, but multiple states only in the most appropriate places
No code generation

manuel-plavsic: > Yeah I'm thinking about family but not in the same way as riverpod, let me show you some pseudo code. Let's say that we have a TodoBloc that handles a single todo item. To work with that item (e.g. retrieve it from the server) it needs a `todoId`
> 
> ```dart
> TodoBloc(this.todoId);
> ```
> 
> The provider would be
> 
> ```dart
> final todoBlocProvider = Provider<TodoBloc>(
>   create: (todoId) => TodoBloc(todoId),
>   dispose: (bloc) => bloc.dispose(),
> );
> ```
> 
> When passed to the ProviderScope widget this would be
> 
> ```dart
> ProviderScope(
>   providers: [
>     todoBlocProvider.withArguments(todoId: "todo_1")
>   ],
> ```
> 
> I'm not sure this is feasible but I remember something like this on mobx, [look here for the reference](https://github.com/mobxjs/mobx.dart/blob/main/mobx/lib/src/core/action.dart#L47-L63)
> 
> In this way we can just wrap the create function provided by the user with the arguments. The method `withArguments` would return `this`, and before returning would store the args.
> 
> When the real create is needed, we can have a private create that wraps the function to make it working

Actually, this might not necessarily be what the library users want. Maybe they want to give a special initial argument, and not all the fields of the provided class. So I believe that the `ProviderWithArg` approach is more flexible.

manuel-plavsic: I removed the maps in the last commit.

manuel-plavsic: reverted

nank1ro:
merging it for now, I'll refactor the code in another PR

PR #104, opened 2025-01-12
Create Maybe types

manuel-plavsic:
This PR improves the error handling and simplifies the API.

It forces the user to consider both cases where a value is not available yet.

For signals, it treats nullable types the same as non-nullable ones: previously the value `null` was unclear and `hasPreviousValue` was necessary with nullable values. With this PR, the API really feels the same both for nullable and non-nullable types.

For providers, it forces the user to handle providers that are not found (IMO this is better than using `solidart_lint` to highlight `get`, `observe` and `update` without the `maybe` in front). Since we can't exclude the possibility of a value contained in a Provider to be null, this PR ensure the difference between the `null` value and a provider not found.

Note that, in both cases, it is possible to bypass the error handling by using `unwrap()`.

Theoretically, an `Option` (or `Maybe`) class could be defined in the `solidart` package and that class could be used instead of having both `MaybeProvidedValue` and `MaybePreviousValue`, since Dart does not really have a ubiquitous `Option` or `Maybe` type. However, I think that these two specific types might make it simpler for new users to understand.

### DIFF
--- a/packages/flutter_solidart/lib/src/models/provider_extensions.dart
+++ b/packages/flutter_solidart/lib/src/models/provider_extensions.dart
@@ -7,14 +7,7 @@ part of '../widgets/provider_scope.dart';
 /// Get the value of a provider.
 extension GetProviderExtension<T> on Provider<T> {
   /// {@macro provider-scope.get}
-  T get(BuildContext context) {
-    final provider = maybeGet(context);
-    if (provider == null) throw ProviderError<T>(this);
-    return provider;
-  }
-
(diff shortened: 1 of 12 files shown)


manuel-plavsic:
NB: I would like to merge this into your [refactor/provider-scope](https://github.com/nank1ro/solidart/tree/refactor/provider-scope) branch and not `dev`.

manuel-plavsic:
It is totally understandable. I am closing here.

PR #106, opened 2025-01-18
refactor: ref #95

medz:
ref: #95


medz:
see [I don't know what I did wrong, the example runs fine ~ the benchmark doesn't run.](https://github.com/nank1ro/solidart/tree/refactor/reactive-v2-part)

PR #122, opened 2025-07-31
Update alien signals to v0.5 & fixed performance regression

### LINKED ISSUE #120: chore: Upgrade dep `alien_signals` to v5 version
> This is a to-do.

Enabling `prefer-inline` by default in upstream dependencies ( `alien_signals: >= 0.4 < 0.5`) will improve performance, but this will increase the user's final code size (approximately 10kb).

This has been resolved in v0.5. 0.5 no longer uses `prefer-inline` and achieves a higher performance level, only slightly lower than the inline version 0.3, but higher than the inline version v0.4.*.

medz:
fixes #120

- replace magic numbers with documented reactive flags
- refactor effect queue to use linked list instead of map

### DIFF
--- a/packages/solidart/lib/src/core/alien.dart
+++ b/packages/solidart/lib/src/core/alien.dart
@@ -1,9 +1,8 @@
 part of 'core.dart';
 
 class _AlienComputed<T> extends alien.ReactiveNode implements _AlienUpdatable {
-  // flags: ReactiveFlags.mutable | ReactiveFlags.dirty
   _AlienComputed(this.parent, this.getter)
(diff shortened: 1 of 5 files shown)


nank1ro:
I really thought it was for performance reasons, although I also thought that saving them as constants was a good practice anyway, also to make the code more readable.

BTW feel free to merge it yourself

medz:
It has no impact on external APIs, it's just an internal optimization of Solidart, so my suggestion is to just release a minor version.

nank1ro:
@all-contributors please add @medz for code, bug, doc

PR #165, opened 2025-12-04
Add authentication flow example

hareshgediya:
This PR introduced the authentication flow example with GoRouter(main.router.dart), and Navigator(main.dart). To demonstrate the use of solidart with flutter.

## Summary by CodeRabbit

* **New Features**
  * Added an auth_flow example with login/logout, persistent session state, guarded routing (auto-redirect to login), a profile page showing account details, and an interactive home screen with a counter.

* **Documentation**
  * Added a README and analysis/configuration for the example.

* **Chores**
  * Added project metadata and ignore rules and included the example in the workspace.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


examples/auth_flow/lib/main.dart
```diff
@@ -0,0 +1,66 @@
...
+        return MaterialApp(
+          navigatorKey: MyApp.navigatorKey,
+          title: 'Auth Demo',
+          debugShowCheckedModeBanner: false,
+          theme: ThemeData(colorScheme: ColorScheme.fromSeed(seedColor: Colors.deepPurple)),
+          home: controller.isLoggedIn.value ? const HomePage(title: 'Home') : const LoginPage(),
+        );
+      },
+    );
+  }
```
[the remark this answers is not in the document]
hareshgediya: 2. Effect callback signature mismatch, this is wrong review. Here is documentation of Effect.
/// Whenever you want to stop the effect from running, you just have to call
/// the returned callback of the `Effect` method:
/// ```dart
/// final disposeEffect = Effect(() { /* your code */ });
/// // later
/// disposeEffect(); // this will stop the effect from running
/// ```

nank1ro:
Thanks for the PR! 🙏 I'll review it as soon as I have some free time

examples/auth_flow/lib/main.dart
nank1ro, line 1: I would remove completely this main example, because if the example is meant to use go_router, than using the Material Navigator should be in a different example.
I don't even know how much this example scales, because you're handling only the authentication, but if there is something more, like user roles, it will become pretty hard to maintain.

examples/auth_flow/lib/main.router.dart
```diff
@@ -0,0 +1,64 @@
...
+import 'package:disco/disco.dart';
+import 'package:flutter/material.dart';
+import 'package:go_router/go_router.dart';
+import 'package:shared_preferences/shared_preferences.dart';
+
+void main() async {
+  WidgetsFlutterBinding.ensureInitialized();
+  final prefs = await SharedPreferences.getInstance();
+
+  runApp(ProviderScope(providers: [sharedPreferenceProvider(prefs)], child: const MyApp()));
```
nank1ro: What you did here is perfectly fine.
You made SharedPreferences synchronous by passing the initialized instance.

However I'm not a fan of SharedPreferences, and using it as a provider is not very convenient.
I would suggest using [localstorage](https://pub.dev/packages/localstorage) instead, which still requires to initialize the object once. But then you can access `localStorage` from anywhere you need in your app, without using a provider.

examples/auth_flow/lib/notifiers/auth_notifier.dart
nank1ro, line 1: What you did here works, but you have to keep in sync both the preferences and the `user` signal which is error prone.

Instead I would use something like this
```dart
/// A signal that holds the user data and manages persistence.
class UserSignal extends Signal<User?> {
  UserSignal()
    : super(
        localStorage.getItem('user') != null
            ? User.fromJson(localStorage.getItem('user')!)
            : null,
      );

  @override
  set value(User? newValue) {
    super.value = newValue;
    if (newValue == null) return localStorage.removeItem('user');
    localStorage.setItem('user', newValue.toJson());
  }

}
```

I haven't tested the code above but it should give you the idea.
In this way, the source of truth is the local storage.
Every time you update the signal value, it will automatically be stored

nank1ro (changes requested):

nank1ro:
I'm not even a fan of go_router. But for the simplicity of this example is pretty fine.
Things tends to be a bit more complicated on a complex project, and I prefer suggesting auto_route instead.

examples/auth_flow/lib/main.dart
hareshgediya: I completely agree that it will not scale, here I just want to demonstrate different approaches, how easily handle auth flow with solidart (with go_router as well). I was also planning to add auto_route example with same codebase so user can get idea about it.

hareshgediya:
> I'm not even a fan of go_router. But for the simplicity of this example is pretty fine. Things tends to be a bit more complicated on a complex project, and I prefer suggesting auto_route instead.

I also prefer auto_route over go_router. Here, I'm using go_router for demonstration because developers usually check go_router first (since it's maintained by the Flutter team). However, I'm also planning to add an auto_route example to this.

examples/auth_flow/lib/main.dart
```diff
@@ -0,0 +1,70 @@
...
+    },
+    routes: [
+      GoRoute(
+        path: '/',
+        builder: (_, _) => const HomePage(title: 'Home'),
+        routes: [GoRoute(path: 'profile', builder: (_, _) => const ProfilePage())],
+      ),
+      GoRoute(path: '/login', builder: (_, _) => const LoginPage()),
+    ],
+  );
```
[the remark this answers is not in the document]
hareshgediya: Thanks for the detailed analysis.

In this codebase, Computed does implement ValueListenable, so it fully satisfies go_router v17's refreshListenable requirement.

Specifically: 

- Computed<T> extends solidart.Computed<T> with ValueListenableSignalMixin<T>
- ValueListenableSignalMixin<T> implements ValueListenable<T>
- ValueListenable extends Listenable, providing addListener / removeListener
- Internally, listener registration is bridged via observe(...), ensuring router refreshes on signal changes

As a result, passing auth.isLoggedIn to GoRouter.refreshListenable is type-safe and works correctly at runtime.
No additional ChangeNotifier wrapper is required in this setup.

examples/auth_flow/lib/main.router.dart
hareshgediya: Thanks for the suggestion - I've removed SharedPreferences and switched to **localstorage**.

examples/auth_flow/lib/domain/user.dart
```diff
@@ -0,0 +1,15 @@
...
+
+@immutable
+class User {
+  const User({required this.id, required this.name, required this.email});
+
+  final String id;
+  final String name;
+  final String email;
+
+  factory User.fromJson(Map<String, dynamic> json) =>
```
nank1ro: I would call it fromMap

examples/auth_flow/lib/domain/user.dart
```diff
@@ -0,0 +1,15 @@
...
+      User(id: json['id'] as String, name: json['name'] as String, email: json['email'] as String);
+
+  Map<String, dynamic> toJson() => {'id': id, 'name': name, 'email': email};
```
nank1ro: I would call it toMap

examples/auth_flow/lib/ui/profile_page.dart
```diff
@@ -0,0 +1,41 @@
...
+import 'package:flutter/material.dart';
+
+class ProfilePage extends StatelessWidget {
+  const ProfilePage({super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    final user = AuthNotifier.provider.of(context).currentUser.value;
+
+    if (user == null) {
```
nank1ro: You're not reacting to the user state with a SignalBuilder

examples/auth_flow/lib/main.dart
```diff
@@ -0,0 +1,76 @@
...
+    return MaterialApp.router(
+      title: 'Auth Demo - GoRouter',
+      debugShowCheckedModeBanner: false,
+      theme: ThemeData(colorScheme: ColorScheme.fromSeed(seedColor: Colors.deepPurple)),
+      routerConfig: router,
+    );
+  }
+}
+
+class AppRouter {
```
nank1ro: What about placing the router in another file?

nank1ro (changes requested):

nank1ro (approved):
LGTM! thanks for the PR 🙏

PR #166, opened 2025-12-06
[WIP] - 3.0: Remove complex custom Reactive Systems and standardize the public API to make it easy to maintain with clear responsibilities and logic. (Goal: Single-file pure code under 1000 lines)

### LINKED ISSUE #156: Upgrade alien_signals to version 2.0
resolves #148

alien_signals uses a more reasonable API layering and offers a slight performance improvement. Memory stability has also been improved.

For Solidart, the biggest benefit is that ReactiveMap/ReactiveList/ReactiveSet can be triggered directly without much additional wrapping.

## Compared to 0.x

- Performance has been significantly improved, with an improvement of approximately 10% in typical scenarios.
- Better preset implementation, allowing users to forgo maintaining their own ReactiveSystem.
- More stable and lower memory usage

## Compared to 1.x

- More logical export layering, eliminating pollution caused by preset and default exports.
- Expose all built-in implementations to avoid writing duplicate code for the same requirements.

# Expected benefits for Solidart:

- Benchmark tests should be able to achieve sub-4-second performance
- Significantly reduce Solidart maintenance costs and complexity.

medz:
1. Remove complex custom Reactive Systems.
2. Migrate to alien_signals 2.1 (version 2.1 has undergone a refactoring to meet Solidart requirements).
3. Create more clearly defined public APIs, reducing incorrect naming conventions like `hasValue` that mark `Signal` as late.
4. Ensure API compatibility and that all current tests pass.

---

resolves #156 ，next PR #167


medz:
@nank1ro I'm wondering if _AlienComputed, _AlienEffect, and _AlienSignal are necessary? What was your original reason for adding them? Shouldn't ReactiveNode be implemented directly in Effect? ​​And shouldn't ComputedNode be implemented directly in Computed, and SignalNode be implemented directly in SignalNode?

nank1ro:
> if _AlienComputed, _AlienEffect, and _AlienSignal are necessary? What was your original reason for adding them? Shouldn't ReactiveNode be implemented directly in Effect? ​​And shouldn't ComputedNode be implemented directly in Computed, and SignalNode be implemented directly in SignalNode?

I don't really remember, maybe for the comparison in update or the disposal?
If it's worth, I would prefer removing anything that is no longer needed

medz:
@nank1ro Both Signal and Computed implement SignalBase. Do you have any suggestions on this?

nank1ro:
> @nank1ro Both Signal and Computed implement SignalBase. Do you have any suggestions on this?

A want to remove it completely if possible and rely on ReadSignal as the base any signal extends.
I even have ReadableSignal which was I mistake I made

medz:
> > @nank1ro Both Signal and Computed implement SignalBase. Do you have any suggestions on this?
> 
> A want to remove it completely if possible and rely on ReadSignal as the base any signal extends. I even have ReadableSignal which was I mistake I made

Perhaps I should reorganize Solidart's base classes (and follow the principle that everything should be public, as you'd expect); I always feel it's a bit disorganized whenever I make changes 😂.

--- Of course, this is on the premise of not changing the current public API.

---
> Besides SignalBase and ReadSignal, I think modifying ReadSignal to the ReadonlySignal interface is a good choice. It standardizes the reading behavior and API of signals (so that it is compatible with both signals and computed signals).

nank1ro:
Not sure if I like "ReadonlySignal" as a name, it's clear but I think even "ReadSignal" is clear and even shorter.
For the rest I agree it should be the base.

The purpose of it is that it has no public value setter, and other signals can be casted to ReadSignal to prevent altering the value.

nank1ro:
> Of course, this is on the premise of not changing the current public API.

Thanks I really appreciate it! 💙☺️
The only thing I think I'll change is autodispose, it's a foot gun for many new users and I'll probably disable it by default and let users to opt-in.
In this way, they can opt-in only after knowing how it works.
But at the end it's just a different default value in `SolidartConfig`

medz:
> > Of course, this is on the premise of not changing the current public API.
> 
> Thanks I really appreciate it! 💙☺️ The only thing I think I'll change is autodispose, it's a foot gun for many new users and I'll probably disable it by default and let users to opt-in. In this way, they can opt-in only after knowing how it works. But at the end it's just a different default value in `SolidartConfig`

If the public Signal API remains unchanged, I believe that such modifications would warrant a release of version 3.0, which seems to contradict the original intention of only supporting certain aspects.

The reasons are as follows:

1. The SignalBase base class has been removed.
2. The ReadonlySignal or ReadSignal interface has been changed.
3. The ReadableSignal base class has been removed.
4. Signal/Computed/Effect exposes the raw ReactiveNode property.

Of course, we could also force a release of version 2.9.0, which currently seems safer because there are no downstream scenarios using SignalBase in Solidart.

nank1ro:
@medz I agree, there are some important breaking changes like removing SignalBase and updating the default SolidartConfig.autoDispose  setting so a new major (3.0) is required

medz:
@nank1ro I'm gradually recreating Solidart's functionality step-by-step from primitives in a single v3.dart file. Currently, the Reactive primitives and auto-dispose functionality (test files added) are complete.

I'd appreciate feedback. Do you have any other comments or suggestions regarding the API design (based on the current functionality of v3.dart)?

Please review this. Thank you! 🙏

medz:
Next PR 👉 #167

PR #167, opened 2025-12-08
Solidart v3: Refactor v3 core API and align ecosystem packages

### LINKED ISSUE #156: Upgrade alien_signals to version 2.0

### LINKED ISSUE #162: [bug] sometimes a computed doesn't autodispose
See /Users/ale/github/cisalpino/lib/features/db/notifier.dart and remove `selectedDBId` to reproduce.

### LINKED ISSUE #168: [BUG] -delayed effect callback caused incorrect dependencies.
Effect design supports a `delay` parameter, which specifies how long the callback should be delayed before execution.

However, this introduces a thread safety issue in synchronous systems. Before a function runs, a reactive system immediately sets the context to the current effect so that dependencies can collect subscriber information.

But the situation changes with delayed callbacks because they depend on a Timer, and Dart cannot truly block threads. The context is reset immediately after being set. Therefore, dependencies in the callback cannot correctly establish dependencies with the current effect.

medz:
before PR：#166, resolves #156 

## Summary by CodeRabbit

* **New Features**
  * v3 public surface: ReadonlySignal, LazySignal, toReadonly(), toValueNotifier() interop; Rich Resource API with resolve/untilReady and state helpers.
  * New reactive collections: ListSignal, MapSignal, SetSignal.
  * Observer API for signal lifecycle.

* **Bug Fixes & Improvements**
  * Simplified Effect API and improved widget builders (SignalBuilder, Show); signals implement Listenable; prefer direct .value assignments.

* **Chores**
  * Version bumps across packages.

* **Tests**
  * Expanded coverage for signals, effects, resources, collections, devtools.


medz:
- [ ] https://github.com/nank1ro/solidart/issues/168

nank1ro:
@medz let me know when the PR is ready to be reviewed 🙏💙

medz:
@nank1ro It's still in draft form. I'll update you when it's ready.

--- PS: I got so frustrated yesterday with the OpenAI GPT-5.1-Codex model that I didn't write any code all day to calm myself down 😂

medz:
@nank1ro I apologize for the delay. I've been quite busy with work lately and may have to wait a while to finish this PR. The core functionality is basically complete. Only some Solidart-specific features and devtools are still incomplete. Thank you for your patience.

nank1ro:
@medz no worries, take all the time you need 💙
i'm overworked of work like you in this last quarter 😅

Feel free to ping me for any issue related to specific features of Solidart

medz:
@nank1ro This PR is ready for review. 🙌

medz:
@nank1ro Tests and checks are green:\n- flutter analyze packages\n- flutter test packages\n- flutter test packages --coverage -r github --fail-fast (overall coverage ~74.73%)

medz:
@coderabbitai

packages/flutter_solidart/lib/src/widgets/signal_builder.dart
```diff
@@ -63,45 +63,57 @@ class SignalBuilder extends StatelessWidget {
...
 
-  late final effect = Effect(
-    scheduler,
-    detach: true,
+  late final Effect _effect = Effect.manual(
+    _runEffect,
     autoDispose: false,
-    autorun: false,
+    detach: true,
   );
```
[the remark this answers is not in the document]
medz: 😂, it looks like you're reviewing outdated code.

examples/toggle_theme/pubspec.yaml
```diff
@@ -34,7 +34,7 @@ dependencies:
   disco: ^1.0.0
   flutter:
     sdk: flutter
-  flutter_solidart: ^2.0.0
+  flutter_solidart: 3.0.0-dev.2
```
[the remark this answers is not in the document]
medz: This wasn't my problem; it was an issue with the example project's code being non-standard 😂. It's fixed now.

packages/solidart/lib/src/solidart.dart
```diff
@@ -0,0 +1,2078 @@
...
+    this.equals = identical,
+    String? name,
+    bool? autoDispose,
+    bool? trackPreviousValue,
+    bool? trackInDevTools,
+  }) : autoDispose = autoDispose ?? SolidartConfig.autoDispose,
+       trackPreviousValue =
+           trackPreviousValue ?? SolidartConfig.trackPreviousValue,
+       trackInDevTools = trackInDevTools ?? SolidartConfig.devToolsEnabled,
+       identifier = ._(name),
```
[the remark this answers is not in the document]
medz: This is the dot abbreviation for Dart 3.10!

packages/solidart/lib/src/solidart.dart
```diff
@@ -0,0 +1,2078 @@
...
+    this.equals = identical,
+    bool? autoDispose,
+    String? name,
+    bool? trackPreviousValue,
+    bool? trackInDevTools,
+  }) : autoDispose = autoDispose ?? SolidartConfig.autoDispose,
+       trackPreviousValue =
+           trackPreviousValue ?? SolidartConfig.trackPreviousValue,
+       trackInDevTools = trackInDevTools ?? SolidartConfig.devToolsEnabled,
+       identifier = ._(name),
```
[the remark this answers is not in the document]
medz: This is the dot abbreviation for Dart 3.10!

packages/solidart/lib/src/solidart.dart
```diff
@@ -0,0 +1,2078 @@
...
+  /// ```
+  ///
+  /// If you want the effect to run immediately, use the [Effect] factory.
+  Effect.manual(
+    VoidCallback callback, {
+    bool? autoDispose,
+    String? name,
+    bool? detach,
+  }) : autoDispose = autoDispose ?? SolidartConfig.autoDispose,
+       identifier = ._(name),
```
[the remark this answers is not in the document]
medz: This is the dot abbreviation for Dart 3.10!

medz:
@nank1ro This PR is ready for review, and I look forward to your feedback.

Compared to the previous version, the code has been reduced by nearly half, while functionality remains almost unchanged. Complete test code has also been rewritten.

There are a few details that require your approval. Should ReadonlySignal be called ReadSignal? I struggled with this for a long time when creating the PR because Read Signal is more like an action than a state.

Also, should ReactiveList/Set/Map be called the old SignalList/Set/Map?

In addition, I removed some APIs inherited from versions 0.x and 1.0 to keep it clean.

I did achieve my goal of keeping the code to around 1000 lines in the draft, but after adding all the comments, the code size increased to around 2000 lines 😂.

nank1ro:
> There are a few details that require your approval. Should ReadonlySignal be called ReadSignal? I struggled with this for a long time when creating the PR because Read Signal is more like an action than a state.

What about ReadableSignal? ReadSignal is shorter, but I agree it could be misleading

> Also, should ReactiveList/Set/Map be called the old SignalList/Set/Map?

Yes, I would not want to break just for a name. `ListSignal` is better IMHO than `ReactiveList`. At the end it's both a `Signal` and a `List`.

> The coverage decreased a lot, by more than 25%

Is it due to missing tests?

medz:
@nank1ro The coverage decrease is because I removed the "ignore test coverage" rule you previously intentionally ignored. Now, there are no rules in the code that ignore coverage (I removed it to maintain true coverage and constantly remind ourselves that we'll need to add coverage tests for devtools, etc., in the future).

Sacrificing readability for brevity isn't worthwhile, especially in development; using readonly is more intuitive.

I think ReactiveList is indeed more conceptual and technical. ListSignal is closer to readability.

Therefore, I will submit a commit later to change ReactiveList/Set/Map to the names you suggested.

nank1ro:
> The coverage decrease is because I removed the "ignore test coverage" rule you previously intentionally ignored. Now, there are no rules in the code that ignore coverage (I removed it to maintain true coverage and constantly remind ourselves that we'll need to add coverage tests for devtools, etc., in the future).

Previously I ignored all the alien_signals coverage because it was already tested in your alien_signals library.

packages/solidart/lib/src/solidart.dart
```diff
@@ -0,0 +1,2103 @@
...
+///
+/// Return `true` when the update should be skipped because values are
+/// considered equivalent.
+typedef ValueComparator<T> = bool Function(T? a, T? b);
+
+/// Lazily produces a value.
+typedef ValueGetter<T> = T Function();
+
+/// A callback that returns no value.
+typedef VoidCallback = ValueGetter<void>;
```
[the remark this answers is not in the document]
medz: @coderabbitai Solidart is a pure native Dart library and does not have the ability to import dart:ui, etc.

medz:
@nank1ro Done! Coverage increased to 100%. All tests passed. The only issue was that data actually sent to devtools couldn't be tested with coverage ignore flags. Also, due to system limitations, abstract classes (const) couldn't be ignored even when tested.

medz:
In summary:

1. This PR fully and correctly implements the autoDispose feature.
2. It cleans up legacy APIs from the 0.x era.
3. It significantly reduces maintenance difficulty, and all design elements are clearly visible.

docs-v2/src/content/docs/advanced/automatic_disposal.mdx
```diff
@@ -6,108 +6,57 @@ sidebar:
 ---
 import { Aside } from '@astrojs/starlight/components';
 
-<Aside type="tip">**TL;DR**: Always dispose Effects and Observations, the rest will be disposed automatically.</Aside>
+<Aside type="tip">
+TL;DR:
+- Dispose Effects you create in widgets.
```
nank1ro:
```suggestion
- Dispose all Effects manually
```

docs-v2/src/content/docs/advanced/automatic_disposal.mdx
```diff
@@ -6,108 +6,57 @@ sidebar:
...
+final counter = Signal(0, autoDispose: true);
 ```
 
-The automatic disposal happens automatically when there are no longer subscribers and listeners (for `Signal`, `Computed`, `Resource`) and when the __currently__ tracked dependencies are all disposed (for `Effect`).
+When auto-dispose is enabled, `Signal`, `Computed`, and `Resource` will dispose
+when they have no subscribers. `Effect` will dispose only after all of its
+tracked dependencies are disposed (not when any single dependency is disposed).
 
-There is a single case that the automatic disposal won't cover:
+## Disposing effects in widgets
```
nank1ro: Effects can be created in business logic classes, it has nothing to do with widgets.
For example, I use `Notifier` classes with `disco` and create/dispose effects from the class itself.

docs-v2/src/content/docs/advanced/automatic_disposal.mdx
```diff
@@ -6,108 +6,57 @@ sidebar:
...
-In fact in the source code the operation is skipped if the object is already disposed:
+Effects created inside widgets should be disposed in `dispose()`:
 
 ```dart
-@override
-void dispose() {
-  // ignore if already disposed
-  if (_disposed) return;
-  ...
+class ExampleState extends State<Example> {
```
nank1ro:
```suggestion
class _ExampleState extends State<Example> {
```

docs-v2/src/content/docs/advanced/automatic_disposal.mdx
```diff
@@ -6,108 +6,57 @@ sidebar:
...
+  late final Effect _effect;
```
nank1ro:
```suggestion
  late final Effect effect;
```

docs-v2/src/content/docs/advanced/automatic_disposal.mdx
```diff
@@ -6,108 +6,57 @@ sidebar:
...
+  final count = Signal(0);
+
+  @override
+  void initState() {
+    super.initState();
+    _effect = Effect(() {
```
nank1ro:
```suggestion
    effect = Effect(() {
```

docs-v2/src/content/docs/advanced/automatic_disposal.mdx
```diff
@@ -6,108 +6,57 @@ sidebar:
...
+      print('count: ${count.value}');
+    });
+  }
+
+  @override
+  void dispose() {
+    _effect.dispose();
```
nank1ro:
```suggestion
    effect.dispose();
```

docs-v2/src/content/docs/advanced/list_signal.mdx
```diff
@@ -5,17 +5,17 @@ sidebar:
...
 
-A `ListSignal` is a special signal that can be used to represent a list of values.
+A `ListSignal` is a reactive wrapper around a `List` that copies on write.
 
 When you modify any element on the list, the `ListSignal` will automatically notify its observers.
 
 ```dart
 final items = ListSignal([1, 2]);
 
-items.observe((previousValue, value) {
```
nank1ro, line 15: observe was special.
It won't fire during creation, unless `fireImmediately` was true.
It is often used to show a Toast/Dialog as a side effect, only tracking the next updates and skipping the current one.

docs-v2/src/content/docs/advanced/map_signal.mdx
```diff
@@ -1,21 +1,44 @@
...
+final effect = Effect(() {
+  print(items.value);
 });
 
-items.add({'c': 3}); // prints "Items changed: {'a': 1, 'b': 2} -> {'a': 1, 'b': 2, 'c': 3}"
-items.remove('a'); // prints "Items changed: {'a': 1, 'b': 2, 'c': 3} -> {'b': 2, 'c': 3}"
+// Later
+effect.dispose();
+items.dispose();
 ```
```
nank1ro: I would not place this paragrah here, it's all described in the automatic_disposal docs.

docs-v2/src/content/docs/advanced/set_signal.mdx
```diff
@@ -2,20 +2,20 @@
 title: Set Signal
 description: Learn SetSignal in solidart
 sidebar:
-  order: 1
+  order: 2
```
nank1ro: you can keep the order as it is, it will be ordered alphabetically (for docs with the same order)

docs-v2/src/content/docs/flutter/show.mdx
```diff
@@ -48,7 +48,7 @@ class _SampleState extends State<Sample> {
   @override
   Widget build(BuildContext context) {
     return Show(
-      when: loggedIn,
+      when: () => loggedIn.value,
```
nank1ro: I would keep `call()` inside signals, because it is very convenient for the Show widget.

docs-v2/src/content/docs/flutter/signal_builder.mdx
```diff
@@ -5,16 +5,14 @@ sidebar:
   order: 1
 ---
 
-import { Aside } from '@astrojs/starlight/components';
+A magic widget builder that automatically rebuilds every time a signal used inside its builder changes.
```
nank1ro:
```suggestion
A magic widget builder that automatically rebuilds every time signals used inside its builder change.
```

docs-v2/src/content/docs/flutter/signal_builder.mdx
```diff
@@ -38,9 +36,5 @@ class _SampleCounterState extends State<SampleCounter> {
 }
 ```
 
-<Aside type="caution">
-By default, if no __active__ signals are detected inside the builder method, an assertion error will be thrown in debug mode.
-This is to help catch mistakes where you might have forgotten to use an __active__ signal inside the builder.
-
-There may be cases where you want to disable this behavior, for example if you want it to build with the latest value of a disposed signal, in that case you can use `SolidartConfig.assertSignalBuilderWithoutDependencies = false` before `runApp`.
```
nank1ro, line 45: Why do you removed the assert?
It was very convenient to discover signals that are incorrectly disposed (when autoDispose is true)

docs-v2/src/content/docs/learning/computed.mdx
```diff
@@ -56,30 +56,30 @@ class User {
...
 
 // adding an effect to print the age
 Effect(() {
-  print('age changed from ${age.previousValue} into ${age.value}');
+  print('age changed to ${age.value}');
 });
 
 // just update the name, the effect above doesn't run because the age has not changed
-user.updateValue((value) => value.copyWith(name: 'new-name'));
+user.value = user.value.copyWith(name: 'new-name');
```
nank1ro: updateValue was giving the untracked value of the signal, this change is not equal to before.

nank1ro (changes requested):
I haven't reviewed all yet, because I found some changes I would like to see updated before continuing to review.

docs-v2/src/content/docs/advanced/list_signal.mdx
nank1ro: We may still use Effect in favor of observe, but we may need a new parameter to skip the first effect reaction.
Because requiring the user to skip the first reaction is not so easy to implement (from the user perspective) because it requires a new boolean variable.

medz:
Hey @nank1ro, regarding the documentation, could you please commit directly? As you've probably noticed, my English is terrible, so I'm not very good at writing docs 😂.

docs-v2/src/content/docs/flutter/signal_builder.mdx
medz: Because I fixed the manual dispose behavior, for example, if a SignalBuilder depends on a Signal, and `a.dispose()` is called, the relationship between SignalBuilder and `a` is broken. At this point, `a` cannot be set to a new value. Furthermore, since `a` and SignalBuilder are no longer dependent after dispose, if `a` continues to be consumed by other subscribers without being disposed to update its value, the SignalBuilder will not be triggered.

The situation you encountered before was simply a side effect caused by an implementation error in the automatic destruction mechanism.

docs-v2/src/content/docs/advanced/automatic_disposal.mdx
```diff
@@ -6,108 +6,57 @@ sidebar:
 ---
 import { Aside } from '@astrojs/starlight/components';
 
-<Aside type="tip">**TL;DR**: Always dispose Effects and Observations, the rest will be disposed automatically.</Aside>
+<Aside type="tip">
+TL;DR:
+- Dispose all Effects manually
+- Signals/Computeds/Resources only auto-dispose when autoDispose is enabled and nothing depends on them.
+- When SolidartConfig.autoDispose is false, you MUST manually dispose Signals, Computeds, and Resources to avoid leaks.
+</Aside>
```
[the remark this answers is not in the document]
medz: @nank1ro After making the changes based on your suggestions, @coderabbitai resubmitted the review comments. Hmm...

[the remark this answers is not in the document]
medz: @coderabbitai In version 3, automatic and manual dispose are as follows:

Automatic Dispose: You can check the test file. When node 'a' is depended on by nodes 'b' and 'c', automatic dispose is enabled in both cases. When 'b' is automatically disposed of, because 'a' is depended on by 'c', and 'c' is still active, only 'b' itself is disconnected from 'a'.

Manual Dispose: Again, using 'a' as an example of nodes depended on by 'b' and 'c', after 'b' is disposed of, 'a' and 'c' are still active, so only 'b' is disconnected. If 'a' is disposed of, and 'b' and 'c' have no other dependencies or their dependencies can be safely disposed of, then all nodes with `autoDispose` enabled will be disconnected.

packages/solidart/lib/src/solidart.dart
```diff
@@ -0,0 +1,2129 @@
...
+
+  @override
+  int get length => value.length;
+
+  @override
+  set length(int newLength) {
+    final current = untrackedValue;
+    if (current.length == newLength) return;
+    value = List<E>.of(current)..length = newLength;
+  }
```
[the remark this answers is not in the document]
medz: This is a known limitation of the Dart SDK's ListMixin, and we cannot resolve it! This issue persists even when using List! This is not a problem we can fix!

docs-v2/src/content/docs/advanced/automatic_disposal.mdx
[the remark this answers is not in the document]
nank1ro: Screenshot 2026-01-12 at 16 20 34

This was correctly mentioned in the old docs.
I suggested to always dispose Effects manually, because it requires less cognitive thinking.
You can dispose a single Effect, and don't have to dispose all dependencies. So disposing a single thing is better than disposing multiple things.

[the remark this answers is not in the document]
nank1ro: > `Effect` auto-disposes only after ALL its tracked dependencies are disposed or disconnected

If the user calls `effect.dispose` it should always dispose, even it tracked dependencies are still alive.

docs-v2/src/content/docs/flutter/signal_builder.mdx
medz: Rewritten and added back

medz:
@nank1ro I've made some changes, please continue to review ❤️

packages/flutter_solidart/lib/src/core/value_listenable_signal_mixin.dart
nank1ro, line 1: There was a reason why there was this mixin.
Every reactive object in flutter_solidart must extend ValueNotifier/ChangeNotifier.
This is especially needed for the `solidart_hooks` package, to be able to use the `useListenable` hook.

packages/solidart_devtools_extension/lib/main.dart
```diff
@@ -134,16 +137,19 @@ class _SignalsState extends State<Signals> {
...
           final data = event.extensionData?.data;
           if (data == null) return;
-          switch (event.extensionKind) {
-            case 'ext.solidart.signal.created':
-            case 'ext.solidart.signal.updated':
-            case 'ext.solidart.signal.disposed':
-              signals[data['_id']] = SignalData(
+          final kind = event.extensionKind;
+          switch (kind) {
+            case 'ext.solidart.v3.signal.created':
```
nank1ro: why v3?
the devtools gets packed together with the solidart package when released. There is no need to support different versions

packages/solidart_hooks/example/lib/pages/use_reactive_list.dart
nank1ro, line 1: rename as before

packages/solidart_hooks/example/lib/pages/use_reactive_map.dart
nank1ro, line 1: rename as before

packages/solidart_hooks/example/lib/pages/use_reactive_set.dart
nank1ro, line 1: rename as before

packages/solidart/lib/src/solidart.dart
```diff
@@ -0,0 +1,2129 @@
...
+  /// final effect = Effect.manual(() {
+  ///   print('count: ${count.value}');
+  /// });
+  ///
+  /// count.value = 1; // no output yet
+  /// effect.run();    // prints "count: 1" and starts tracking
+  /// ```
+  ///
+  /// If you want the effect to run immediately, use the [Effect] factory.
+  Effect.manual(
```
nank1ro: why not calling it `lazy` instead of manual?

I still find it a bit inconvenient, given that the typical usage is just "skip the current value, trigger on next reaction". I would like a boolean parameter like "skipCurrent" or "skipFirst" (don't know the best name)/

packages/solidart/lib/src/solidart.dart
```diff
@@ -0,0 +1,2129 @@
...
+    if (isDisposed) return;
+    if (_resolveFuture != null) return _resolveFuture!;
+    if (_resolved) return;
+
+    _resolved = true;
+    _resolveFuture = _doResolve().whenComplete(() {
+      _resolveFuture = null;
+    });
+
+    return _resolveFuture!;
```
nank1ro: can you clarify what are you doing here?
what is `_resolveFuture`, why is needed, and why are you forcing it to be non-nullable?

nank1ro (changes requested):

packages/solidart/lib/src/solidart.dart
medz: `_doResolve` returns a Future, where `whenComplete` will always execute `return _resolveFuture`. At the time of `return`, `_resolveFuture` will definitely exist. Adding `!` is purely for linting purposes. `_resolveFuture` is a property member with a nullable signature (which we need), but it will always be a Future at the time of `return`. Adding an extra variable to receive and return it makes no sense. Adding only one character, `!`, is more cost-effective than adding about 10 characters.

packages/solidart/lib/src/solidart.dart
medz: We discussed this in another PR before. "Lazy" doesn't make sense, and its behavior isn't lazy at all. If you want to skip the first callback run of an effect, you can't collect dependencies. If you only want to control when the effect runs for the first time, the manual is indeed a bit cumbersome. But the purpose of the manual is to provide general and high-level control.

Explicit manual control is more cost-effective than creating requirements.

nank1ro: What about reintroducing observe as an extension like before? It would cover the previous usage users have, and doesn't require the user to collect a new variable (like skipped) themselves and refactor all usages of observe.
https://github.com/nank1ro/solidart/blob/40119f4fcdf1f78d356174bc7cb836633cab3cf8/packages/solidart/lib/src/core/extensions.dart#L9C1-L34C2

If you look closely at the code, it uses a normal Effect, but skips calling the listener on creation.

docs-v2/src/content/docs/advanced/list_signal.mdx
```diff
@@ -2,20 +2,42 @@
...
-  print("Items changed: $previousValue -> $value");
+Effect(() {
+  print('Items: ${items.value}');
 });
```
nank1ro: you added back `observe`, you can revert the example

docs-v2/src/content/docs/advanced/list_signal.mdx
```diff
@@ -2,20 +2,42 @@
...
+```dart
+var skipped = false;
+Effect(() {
+  items.value; // track
+  if (!skipped) {
+    skipped = true;
+    return;
+  }
+  // side effect here
+});
```
nank1ro: you added back observe so this comment is outdated

docs-v2/src/content/docs/advanced/map_signal.mdx
```diff
@@ -1,21 +1,31 @@
...
-When you modify any element on the map, the `MapSignal` will automatically notify its observers.
+When you modify any element on the map, the `MapSignal` will automatically
+notify its observers.
 
 ```dart
+import 'package:solidart/solidart.dart';
+
 final items = MapSignal({'a': 1, 'b': 2});
 
-items.observe((previousValue, value) {
```
nank1ro, line 15: observe is back, revert

docs-v2/src/content/docs/advanced/set_signal.mdx
```diff
@@ -2,20 +2,23 @@
...
+A `SetSignal` is a reactive wrapper around a `Set` that copies on write.
 
 When you modify any element on the set, the `SetSignal` will automatically notify its observers.
 
 ```dart
 final items = SetSignal({1, 2});
 
-items.observe((previousValue, value) {
-  print("Items changed: $previousValue -> $value");
+Effect(() {
```
nank1ro: observe is back, revert

docs-v2/src/content/docs/learning/computed.mdx
```diff
@@ -56,30 +56,30 @@ class User {
...
 
 // adding an effect to print the age
 Effect(() {
-  print('age changed from ${age.previousValue} into ${age.value}');
+  print('age changed to ${age.value}');
 });
 
 // just update the name, the effect above doesn't run because the age has not changed
-user.updateValue((value) => value.copyWith(name: 'new-name'));
+user.value = user.untrackedValue.copyWith(name: 'new-name');
```
nank1ro: there is no need to use untrackedValue here, we're not inside an Effect, this comment is valid for the whole file.

docs-v2/src/content/docs/learning/effects.mdx
```diff
@@ -13,32 +13,47 @@ The effect automatically subscribes to any signal and reruns when any of them ch
...
-disposeEffect();
+effect.dispose();
 ```
 
-<Aside>The effect runs immediately with the current signals' values</Aside>
+<Aside>The effect runs immediately with the current signals' values.</Aside>
 
 <Aside type="caution">An effect is useless after it is disposed, you must not use it anymore.</Aside>
+
+## Lazy effects
```
nank1ro:
```suggestion
## Manual effects
```

docs-v2/src/content/docs/learning/signal.mdx
```diff
@@ -51,39 +51,24 @@ To change the value, you can use:
...
 ```dart
 final counter = Signal(0);
-final readOnlyCounter = counter.toReadSignal();
+final readOnlyCounter = counter.toReadonly();
 print(readOnlyCounter.value); // prints 0
 // The code below is invalid and will cause a compile-time error because no setter is present
 readOnlyCounter.value = 2;
 ```
 
-## Observe the signal
```
nank1ro, line 69: I would not remove this paragrah, it is still true that we have observe, and not only Effect

docs-v2/src/content/docs/migration.mdx
```diff
@@ -0,0 +1,269 @@
...
+### Signals and ReadSignal
+
+| v2 API | v3 API | Notes |
+| --- | --- | --- |
+| `ReadSignal<T>` | `ReadonlySignal<T>` | Read-only interface renamed. Use `toReadonly()` to create one. |
+| `toReadSignal()` | `toReadonly()` | Read-only conversion rename. |
+| `signal()` / `ReadSignal.call()` | `signal.value` | Call operator still works; `.value` is preferred. |
+| `signal.set(value)` | `signal.value = value` | Setter only. |
+| `signal.updateValue(fn)` | `signal.value = fn(signal.untrackedValue)` | Use `untrackedValue` to match v2 semantics. |
+| `signal.observe(listener, fireImmediately)` | `Effect(() { ... })` | Use effects for reactive side effects. |
```
nank1ro: No longer true

docs-v2/src/content/docs/migration.mdx
```diff
@@ -0,0 +1,269 @@
...
+
+## 3) API mapping (v2 -> v3)
+
+### Signals and ReadSignal
+
+| v2 API | v3 API | Notes |
+| --- | --- | --- |
+| `ReadSignal<T>` | `ReadonlySignal<T>` | Read-only interface renamed. Use `toReadonly()` to create one. |
+| `toReadSignal()` | `toReadonly()` | Read-only conversion rename. |
+| `signal()` / `ReadSignal.call()` | `signal.value` | Call operator still works; `.value` is preferred. |
```
nank1ro: this is not a change then

docs-v2/src/content/docs/migration.mdx
```diff
@@ -0,0 +1,269 @@
...
+| `signal.set(value)` | `signal.value = value` | Setter only. |
+| `signal.updateValue(fn)` | `signal.value = fn(signal.untrackedValue)` | Use `untrackedValue` to match v2 semantics. |
+| `signal.observe(listener, fireImmediately)` | `Effect(() { ... })` | Use effects for reactive side effects. |
+| `signal.until(condition, timeout)` | — | Removed. Use `Completer` + `Effect` (example below). |
```
nank1ro: I would not have removed `until` which is pretty useful, given that you can `await` the returned future.

docs-v2/src/content/docs/migration.mdx
```diff
@@ -0,0 +1,269 @@
...
+});
+await completer.future;
+```
+
+### Computed
+
+| v2 API | v3 API | Notes |
+| --- | --- | --- |
+| `Computed(selector, options)` | `Computed(selector, ...)` | Options are named params. |
+| `computed()` | `computed.value` | Call operator still works; `.value` is preferred. |
```
nank1ro: 1. what is this ? lowercase? Only `Computed` should be there
2. This is not a change

docs-v2/src/content/docs/migration.mdx
```diff
@@ -0,0 +1,269 @@
...
+| `Computed.run()` | `computed.value` | No manual run; reading evaluates lazily. |
+
+### Effect
+
+| v2 API | v3 API | Notes |
+| --- | --- | --- |
+| `Effect((dispose) { ... }, onError, options)` | `Effect(() { ... }, autoDispose:, name:, detach:)` | Callback no longer receives `dispose`. |
```
nank1ro: This is already true for the v2

docs-v2/src/content/docs/migration.mdx
```diff
@@ -0,0 +1,269 @@
...
+| `EffectOptions(delay: ...)` | — | Delay removed; use your own debounce/timer if needed. |
+| `effect()` to dispose | `effect.dispose()` | Call operator removed. |
+| `Effect.manual` | `Effect.manual` | Still available; call `run()` to start tracking. |
```
nank1ro: I don't understand the sentence at all

docs-v2/src/content/docs/migration.mdx
```diff
@@ -0,0 +1,269 @@
...
+
+### Resource
+
+| v2 API | v3 API | Notes |
+| --- | --- | --- |
+| `Resource(fetcher, source: SignalBase)` | `Resource(fetcher, source: ReadonlySignal)` | `SignalBase` removed. |
+| `ResourceOptions` | named params | `name`, `autoDispose`, `equals`, `trackInDevTools`, `useRefreshing`, `debounceDelay`, `trackPreviousState`. |
```
nank1ro: this was already true in the v2

docs-v2/src/content/docs/migration.mdx
```diff
@@ -0,0 +1,269 @@
...
+| `resource.update((state) => ...)` | `resource.state = ...` | Write the new state directly. |
+| `resource.select(selector)` | `Computed(() => ...)` | Use `Computed` to derive data. |
+| `resource.untilReady()` | — | Removed. Use `Completer` + `Effect`. |
```
nank1ro: This was pretty useful, why remove it?

docs-v2/src/content/docs/migration.mdx
```diff
@@ -0,0 +1,269 @@
...
+| `ResourceState.on / maybeOn` | `when / maybeWhen` | `on`/`maybeOn` removed. |
```
nank1ro: already valid in the v2

docs-v2/src/content/docs/migration.mdx
```diff
@@ -0,0 +1,269 @@
...
+| `ResourceState.value` / `call()` | `asReady?.value` / `when` | Use extensions and pattern matching. |
```
nank1ro: nothing changed here

docs-v2/src/content/docs/migration.mdx
```diff
@@ -0,0 +1,269 @@
...
+
+### Collections
+
+| v2 API | v3 API | Notes |
+| --- | --- | --- |
+| `ListSignal<T>` | `ListSignal<T>` | Copy-on-write collection. |
+| `SetSignal<T>` | `SetSignal<T>` | Copy-on-write collection. |
+| `MapSignal<K,V>` | `MapSignal<K,V>` | Copy-on-write collection. |
+| `useListSignal` | `useListSignal` | Hook for list signal. |
+| `useSetSignal` | `useSetSignal` | Hook for set signal. |
```
nank1ro: Nothing changed from v2 to v3

docs-v2/src/content/docs/migration.mdx
```diff
@@ -0,0 +1,269 @@
...
+| `useMapSignal` | `useMapSignal` | Hook for map signal. |
+
+### Flutter integration
+
+| v2 API | v3 API | Notes |
+| --- | --- | --- |
+| `ReadSignal.toValueNotifier()` | `ReadonlySignal.toValueNotifier()` | Now provided via extension. |
+| `ValueListenable.toSignal()` | `ValueListenable.toSignal()` | Still available via extension. |
```
nank1ro: nothing changed for the user

docs-v2/src/content/docs/migration.mdx
```diff
@@ -0,0 +1,269 @@
...
+  SolidartConfig.autoDispose = true;
+  final signal = Signal(0, autoDispose: true);
+  ```
+- **DevTools tracking default changed**: v2 defaults to `kDebugMode`, v3 defaults to `false`. Enable it explicitly if you need observer events:
+  ```dart
+  SolidartConfig.devToolsEnabled = true;
+  final signal = Signal(0, trackInDevTools: true);
+  ```
+- **Previous value tracking is read-driven**: `previousValue` / `previousState` update only after a tracked read.
+- **Nullable previous values**: `hasPreviousValue` was removed, so `previousValue == null` can mean either “no previous value” or “previous value was null.” Use a sentinel or wrap nullable values if you need to distinguish.
```
nank1ro: I would have keeped `hasPreviousValue` because it simplify to distinguish

docs-v2/src/content/docs/migration.mdx
```diff
@@ -0,0 +1,269 @@
...
+- **Reactive collections are copy-on-write**: mutating methods replace the internal collection value. Avoid mutating `untrackedValue` directly.
+- **Effect timing and errors**: `EffectOptions.delay` and `onError` were removed. Use `Timer`/debounce and `try/catch` inside the effect.
```
nank1ro: why removing onError?

docs-v2/src/content/docs/migration.mdx
```diff
@@ -0,0 +1,269 @@
...
+- **Nested effects**: v3 adds `detach` / `SolidartConfig.detachEffects` to control whether inner effects become dependencies of their parent.
```
nank1ro: already present in v2

docs-v2/src/content/docs/migration.mdx
```diff
@@ -0,0 +1,269 @@
...
+```
+
+**After (v3):**
+```dart
+Effect(() {
+  final next = counter.value;
+  final prev = counter.untrackedPreviousValue;
+  print('from $prev to $next');
+});
+```
```
nank1ro: observe was added again by you

docs-v2/src/content/docs/migration.mdx
```diff
@@ -0,0 +1,269 @@
...
+
+**Before (v2):**
+```dart
+final todos = ReactiveList<String>([]);
+```
+
+**After (v3):**
+```dart
+final todos = ListSignal<String>([]);
+```
```
nank1ro: surely docs have been written by an AI 😅

docs-v2/src/content/docs/migration.mdx
```diff
@@ -0,0 +1,269 @@
...
+untracked(() {
+  // reads here do not create dependencies
+  print(counter.value);
+});
+
+batch(() {
+  counter.value++;
+  other.value++;
+});
+```
```
nank1ro: already present in v2

docs-v2/src/content/docs/migration.mdx
```diff
@@ -0,0 +1,269 @@
...
+final user = Resource(fetchUser, source: userId, debounceDelay: Duration(milliseconds: 300));
+
+SignalBuilder(
+  builder: (_, __) => user.state.when(
+    ready: (data) => Text(data.name),
+    loading: () => const CircularProgressIndicator(),
+    error: (e, _) => Text('Error: $e'),
+  ),
+);
+```
```
nank1ro: nothing has changed

docs-v2/src/content/docs/migration.mdx
```diff
@@ -0,0 +1,269 @@
...
+```dart
+final notifier = counter.toReadonly().toValueNotifier();
+final signal = someValueListenable.toSignal();
+```
+
+## 10) Removed APIs (summary)
+
+- `toReadSignal`, `ReadSignal`, `SignalBase`
+- `SignalOptions`, `ResourceOptions`, `EffectOptions`
+- `updateValue`, `observe`, `until`
```
nank1ro: `observe` was readded, `until` should be readded

docs-v2/src/content/docs/migration.mdx
```diff
@@ -0,0 +1,269 @@
...
+- `ReactiveList`, `ReactiveMap`, `ReactiveSet`
```
nank1ro: they were never a thing

docs-v2/src/content/docs/migration.mdx
```diff
@@ -0,0 +1,269 @@
...
+- `Debouncer` and other v2-only utils
+
+If you hit a missing API, prefer the v3 primitives (`Signal`, `Computed`,
+`Effect`, `Resource`) and the new helpers (`untracked`, `batch`).
```
nank1ro: no new helpers, already present in v2

examples/auth_flow/pubspec.yaml
```diff
@@ -9,10 +9,11 @@ environment:
 resolution: workspace
 
 dependencies:
+  cupertino_icons: ^1.0.2
```
nank1ro: why cupertino_icons is back?

examples/github_search/pubspec.yaml
```diff
@@ -30,10 +30,11 @@ resolution: workspace
 # the latest version available on pub.dev. To see which dependencies have newer
 # versions available, run `flutter pub outdated`.
 dependencies:
+  cupertino_icons: ^1.0.2
```
nank1ro: why cupertino_icons is back?

examples/infinite_scroll/pubspec.yaml
```diff
@@ -9,10 +9,11 @@ environment:
 resolution: workspace
 
 dependencies:
+  cupertino_icons: ^1.0.2
```
nank1ro: why cupertino_icons is back?

examples/todos/pubspec.yaml
```diff
@@ -9,10 +9,11 @@ environment:
 resolution: workspace
 
 dependencies:
+  cupertino_icons: ^1.0.2
```
nank1ro: why cupertino_icons is back?

examples/toggle_theme/pubspec.yaml
```diff
@@ -31,10 +31,11 @@ resolution: workspace
 # the latest version available on pub.dev. To see which dependencies have newer
 # versions available, run `flutter pub outdated`.
 dependencies:
+  cupertino_icons: ^1.0.2
```
nank1ro: why cupertino_icons is back?

packages/flutter_solidart/example/lib/pages/lazy_counter.dart
```diff
@@ -9,7 +9,7 @@ class LazyCounterPage extends StatefulWidget {
 }
 
 class _LazyCounterPageState extends State<LazyCounterPage> {
-  final counter = Signal<int>.lazy(name: 'lazyCounter');
+  final counter = LazySignal<int>(name: 'lazyCounter');
```
nank1ro: Why do we need a new class when a factory was already pretty simple and explicit?

packages/flutter_solidart/example/lib/pages/lazy_counter.dart
```diff
@@ -18,7 +18,7 @@ class _LazyCounterPageState extends State<LazyCounterPage> {
       body: Center(
         child: SignalBuilder(
           builder: (_, _) {
-            return switch (counter.hasValue) {
```
nank1ro, line 21: why this API change?

packages/flutter_solidart/example/lib/pages/list_signal.dart
```diff
@@ -15,14 +13,6 @@ class ListSignalPage extends StatefulWidget {
...
   final items = ListSignal([1, 2], name: 'items');
 
-  @override
-  void initState() {
-    super.initState();
-    items.observe((previousValue, value) {
-      print("Items changed: $previousValue -> $value");
-    });
-  }
-
```
nank1ro, line 25: revert

packages/flutter_solidart/example/lib/pages/map_signal.dart
```diff
@@ -17,14 +15,6 @@ class MapSignalPage extends StatefulWidget {
...
   final items = MapSignal({'a': 1, 'b': 2}, name: 'items');
 
-  @override
-  void initState() {
-    super.initState();
-    items.observe((previousValue, value) {
-      print("Items changed: $previousValue -> $value");
-    });
-  }
-
```
nank1ro, line 27: revert

packages/flutter_solidart/example/lib/pages/set_signal.dart
```diff
@@ -15,14 +13,6 @@ class SetSignalPage extends StatefulWidget {
...
   final items = SetSignal({1, 2}, name: 'items');
 
-  @override
-  void initState() {
-    super.initState();
-    items.observe((previousValue, value) {
-      print("Items changed: $previousValue -> $value");
-    });
-  }
-
```
nank1ro, line 25: revert

packages/flutter_solidart/example/lib/pages/show.dart
```diff
@@ -20,7 +20,7 @@ class _ShowPageState extends State<ShowPage> {
         actions: [
           TextButton(
             style: TextButton.styleFrom(foregroundColor: Colors.white),
-            onPressed: loggedIn.toggle,
```
nank1ro, line 23: do you think the toggle extension was superfluous?

packages/flutter_solidart/example/lib/main.dart
```diff
@@ -17,24 +17,35 @@ import 'package:flutter_solidart/flutter_solidart.dart';
...
-  void didUpdateSignal(SignalBase<Object?> signal) {
+  void didUpdateSignal(ReadonlySignal<Object?> signal) {
     dev.log(
-      'didUpdateSignal(name: ${signal.name}, previousValue: ${signal.previousValue}, value: ${signal.value})',
+      'didUpdateSignal(name: ${signal.identifier.name}, previousValue: ${signal.previousValue}, value: ${_safeValue(signal)})',
     );
   }
 }
 
+Object? _safeValue(ReadonlySignal<Object?> signal) {
```
nank1ro: why this? 😕

packages/flutter_solidart/CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
+## 3.0.0-dev.2
+
+- **CHORE**: Bump `solidart` dependency to `3.0.0-dev.2`.
+- **DOCS**: Refresh examples for `LazySignal` and v3 dependency pins.
+
+## 3.0.0-dev.1
+
+- **BREAKING**: Remove Flutter-specific core wrappers under `src/core/*`; `flutter_solidart` now re-exports `solidart/solidart.dart` directly.
+- **BREAKING**: Replace `ListSignal`/`MapSignal`/`SetSignal` with `ReactiveList`/`ReactiveMap`/`ReactiveSet`.
```
nank1ro: no longer true

packages/flutter_solidart/CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
+## 3.0.0-dev.2
+
+- **CHORE**: Bump `solidart` dependency to `3.0.0-dev.2`.
+- **DOCS**: Refresh examples for `LazySignal` and v3 dependency pins.
+
+## 3.0.0-dev.1
+
+- **BREAKING**: Remove Flutter-specific core wrappers under `src/core/*`; `flutter_solidart` now re-exports `solidart/solidart.dart` directly.
```
nank1ro: no longer true

packages/flutter_solidart/CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
...
+- **BREAKING**: Replace `ListSignal`/`MapSignal`/`SetSignal` with `ReactiveList`/`ReactiveMap`/`ReactiveSet`.
+- **BREAKING**: Replace `toReadSignal()` with `toReadonly()` and remove legacy `SignalBase`/`ReadableSignal` surfaces.
+- **FEAT**: Add v3 conversion extensions: `ReadonlySignal.toValueNotifier()` and `ValueListenable.toSignal()`.
```
nank1ro: already present in v2

packages/flutter_solidart/CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
...
+- **REFACTOR**: Rewrite `SignalBuilder` and `Show` to use v3 effects + dependency tracking.
```
nank1ro: nothing that affects the user

nank1ro (changes requested):

docs-v2/src/content/docs/migration.mdx
medz: Because the delay has been removed, it can now be controlled manually.

docs-v2/src/content/docs/migration.mdx
medz: yes😂，i'll revert docs

examples/auth_flow/pubspec.yaml
medz: i'll remove

packages/flutter_solidart/CHANGELOG.md
```diff
@@ -1,3 +1,31 @@
...
+- **REFACTOR**: `Show` simplified to a `StatelessWidget` backed by `SignalBuilder`.
+
+### Interop
+
+- **BREAKING**: `SignalBase.toValueNotifier()` becomes `ReadonlySignal.toValueNotifier()`, and `ValueNotifier.toSignal()` becomes `ValueListenable.toSignal()`.
+
+### Dependencies
+
+- **CHORE**: Bump `solidart` to `3.0.0-dev.0`.
+
```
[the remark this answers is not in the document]
medz: alien_signal dependency updates are not important

packages/flutter_solidart/CHANGELOG.md
```diff
@@ -1,3 +1,31 @@
+## 3.0.0-dev.0 (Unreleased)
+
+### Signals
+
+- **BREAKING**: Flutter `Signal`, `Computed`, `Resource`, and collection wrappers now implement `ValueListenable` only (the `ValueNotifier` mixins are removed).
+- **REMOVED**: `ValueNotifierSignalMixin` and `ValueListenableSignalMixin` exports.
+- **BREAKING**: `Signal.toReadSignal()` is replaced by `toReadonly()` and `ReadableSignal` is removed in favor of `ReadonlySignal`.
+- **REMOVED**: `Signal.updateValue` and `ToggleBoolSignal`.
+- **ADDED**: `LazySignal` wrapper returned by `Signal.lazy`.
```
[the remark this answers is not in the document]
medz: The naming convention has been reverted to v2, so no further explanation is needed. Only internal modifications have been made to v3.

packages/solidart_lint/pubspec.yaml
```diff
@@ -27,5 +27,5 @@ dependencies:
 dev_dependencies:
   lints: ^6.0.0
   test: ^1.25.2
-  solidart: ^2.0.0
-  flutter_solidart: ^2.0.0
+  solidart: 3.0.0-dev.0
+  flutter_solidart: 3.0.0-dev.0
```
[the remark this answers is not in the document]
medz: If the version number is not updated in the pub workspace, flutter pub get will fail.

docs-v2/src/content/docs/migration.mdx
```diff
@@ -0,0 +1,169 @@
...
+## 7) Quick migration checklist
+
+- [ ] Upgrade versions in all solidart ecosystem packages
+- [ ] Replace `ReadSignal`/`SignalBase` with `ReadonlySignal`
+- [ ] Replace `toReadSignal()` with `toReadonly()`
+- [ ] Remove `updateValue` usages (use `.value = ...`)
+- [ ] Update Effect usage (no `delay/onError`, dispose via `.dispose()`)
+- [ ] Update Resource state handling (`state =`, `when/maybeWhen`)
+- [ ] Update Flutter extensions (`toValueNotifier`/`toSignal`)
+- [ ] Adjust hooks disposal expectations
```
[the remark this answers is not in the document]
medz: No, it's not necessary. During the early development of v3, List/Set/MapSignal was renamed to ReactiveList/Set/Map. However, it has now been renamed back to the v2 naming convention, so no migration is required.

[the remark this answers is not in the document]
medz: No, it's not necessary. During the early development of v3, List/Set/MapSignal was renamed to ReactiveList/Set/Map. However, it has now been renamed back to the v2 naming convention, so no migration is required.

packages/solidart/lib/src/core/signal.dart
```diff
@@ -1,172 +1,186 @@
...
+  /// {@macro solidart.signal}
+  ///
+  /// This is a lazy signal: it has no value at construction time.
+  /// Reading [value] before the first assignment throws [StateError].
+  factory Signal.lazy({
+    String? name,
+    bool? autoDispose,
+    ValueComparator<T> equals,
+    bool? trackPreviousValue,
+    bool? trackInDevTools,
```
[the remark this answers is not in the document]
medz: You seem unfamiliar with Dart syntax, also known as factories. Parameter signature forwarding = entity. That is, whatever the constructor type signature of `LazySignal` is, `Signal.lazy` must also be the same. `equals` either passes through or is completely omitted!

packages/solidart/lib/src/core/signal.dart
```diff
@@ -1,172 +1,186 @@
...
+    final pending = pendingValue;
+    if (current is Some<T> &&
+        pending is Some<T> &&
+        equals(pending.value, current.value)) {
+      return false;
+    }
+
+    if (trackPreviousValue && current is Some<T>) {
+      _previousValue = current;
+    }
```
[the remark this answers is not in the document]
medz: Can't you see I retracted the v3 documentation? And my previous behavior didn't make sense. I didn't write the v3 documentation! I'm bad at writing documentation! Stop reminding me about the documentation!!!

packages/solidart/lib/src/resources/resource.dart
```diff
@@ -0,0 +1,325 @@
...
+      (data) {
+        if (_isStale(requestId)) return;
+        state = ResourceState<T>.ready(data);
+      },
+      onError: (Object error, StackTrace stackTrace) {
+        if (_isStale(requestId)) return;
+        state = ResourceState<T>.error(error, stackTrace: stackTrace);
+      },
+    );
+  }
```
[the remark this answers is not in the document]
medz: This is still the behavior of v2, and it has a bug. It will now be handled automatically! The test files are there! Go check the tests!!!

packages/solidart/lib/src/core/disposable.dart
```diff
@@ -0,0 +1,90 @@
...
+      }
+    } finally {
+      cleanups.clear();
+    }
+  }
+
+  @mustCallSuper
+  @override
+  void onDispose(VoidCallback callback) {
+    cleanups.add(callback);
```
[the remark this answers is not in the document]
medz: It doesn't matter

packages/solidart/lib/src/resources/resource.dart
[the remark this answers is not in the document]
medz: In addition, _resubscribe is synchronous code! This means that listen cannot emit new events before _resubscribe is called.

docs-v2/src/content/docs/migration.mdx
medz: Restored

packages/flutter_solidart/example/lib/pages/show.dart
medz: The toggle isn't intuitive enough. Furthermore, it only involves omitting two or three extra characters, while the SolidArt code requires exporting a new type. Also, the use case for bool toggling is very low. Therefore, I don't think it's worthwhile to create a special extension for this.

medz:
@coderabbitai Regenerate PR Summary & title

SAGARSURI:
@nank1ro  when is this happening? Looking forward to this migration. Thanks for this great package 🙏🏽

nank1ro:
> when is this happening? Looking forward to this migration. Thanks for this great package 🙏🏽

Hope very soon, I'm working on something else right now, will come back after finishing the current work

nank1ro:
closed in favor of https://github.com/nank1ro/solidart/pull/175

